PR #58, opened 2014-01-25
Reject an invalid date, like February 31

tevanoff:
This will reject an invalid date instead of crashing.

I figured this was the best way to handle this. Let me know what you think. Cheers!


halan:
https://github.com/apotonick/reform/pull/289

apotonick:
Thanks @tevanoff and sorry for not responding!

PR #77, opened 2014-03-25
Make it possible to have a nested form that uses composition.

TylerRick:
For example, `collection :invitations, form: InvitationForm`, where InvitationForm uses composition.

Whenever I try this, I get this error: `relationship delegated to invitation.relationship, but invitation is nil`.

How do we get invitation to *not* be nil?

### DIFF
--- a/test/nested_form_test.rb
+++ b/test/nested_form_test.rb
@@ -166,6 +166,49 @@ class NestedFormTest < MiniTest::Spec
     }
   end
 
+
+  class NestedFormWithCompositionTest < MiniTest::Spec
+    class InvitationForm < Reform::Form
+      include Composition
+      model :invitation
+      property :relationship, on: :invitation
+      property :name,         on: :recipient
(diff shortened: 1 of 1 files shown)


contentfree:
Ran into this today. This seems like a very common use of form reuse. The example in the docs puts a collection of songs in the module and then includes that, which seems odd… It seems like the intuitive behavior would be to create a form for Song then in an AlbumForm create a collection that reuses the SongForm. 

(Maybe that works today, but I can't test that because I'm using composition and thus ran into this issue.)

Is there a way to use a modular form that uses Composition in a collection in another form yet?

apotonick:
@contentfree Can you make an example of how your data model and the form configuration would look like? How would the nested composition collection know how to find the contained elements?

fran-worley:
Closing this as it's not been updated for rather a long time. Feel free to reopen if you need!

PR #145, opened 2014-09-29
Undefined method 'has_attribute?' from simple form

bookwyrm:
SimpleForm expects a model to respond to has_attribute? while building a form in the view. I just updated from Reform v1.0.4 to v1.1.1 and SimpleForm started throwing and undefined method error.

This PR adds the behavior back in to appease SimpleForm.


fran-worley:
Closing in favour of #176

PR #177, opened 2014-12-04
Delegate reflect_on_association to the model class to support simple_form's association input

### LINKED ISSUE #122: Using the simple_form association form helper doesn't work?
If I try to use the `association` helper in a simple_form, I get an `Association not found` error, even if the underlying AR model has that association.

Example:  

```
= simple_form_for @object do |f|
  = f.association :association_name
```

Looking at the simple_form code [here](https://github.com/plataformatec/simple_form/blob/ed3de1c2b566a962697529e824953f6b280e2867/lib/simple_form/form_builder.rb#L550) they seem to check for the association by calling `reflection_on_association` on the class of the object underlying the form. And it seems that `Reform::Form` does not respond to this method, so the `association` helper fails.

Is there something else I should be doing/specifying in my Form object to make this work, or is it just not supported?

Thanks

zavan:
This fixes #122

Since simple_form relies on `Model.reflect_on_association` to discover model associations when using the association input 
(see https://github.com/plataformatec/simple_form/blob/ed3de1c2b566a962697529e824953f6b280e2867/lib/simple_form/form_builder.rb#L550)
we need to delegate this method from the form class to the model class.

Thanks to @smartygus for spotting this and providing an initial solution for this.

### DIFF
--- a/lib/reform/form/model_reflections.rb
+++ b/lib/reform/form/model_reflections.rb
@@ -13,7 +13,7 @@ module Reform::Form::ModelReflections
     # Delegate reflect_on_association to the model class to support simple_form's
     # association input.
     def reflect_on_association(*args)
-      model_name.name.constantize.reflect_on_association(*args)
+      model_name.to_s.constantize.reflect_on_association(*args)
     end
   end
 


brentonannan:
I'm having trouble with this on Rails 4.0.13 - getting:

    Completed 500 Internal Server Error in 1883ms
    
    SystemStackError - stack level too deep:
      actionpack (4.0.13) lib/action_dispatch/middleware/reloader.rb:70:in `'

Annoyingly, I can't even get to a `binding.pry` placed at the top of `association` in `lib/simple_form/form_builder.rb`, or within `reflect_on_association` in `lib/reform/form/model_reflections.rb` to see what's being called.

Any ideas on where to even start debugging this @apotonick/@zavan?

apotonick:
@brentonannan You can start debugging with `raise` and find out what method is called recursively!

brentonannan:
Because this popped up again with a co-worker, and he found my question here, just thought I'd put down for anyone looking that I ended up just going with:

```
= f.input :accent_id, as: :select, collection: Accent.all, label: I18n.t("attributes.accent")
```

PR #188, opened 2015-01-17
Fix isolated engine namespace

simoniong:
This is a fix to determine the correct route_key for Model defined in namespace-isolated rails engine.

I have a namespace-isolated rails engine(let's say, named `project_core` ), which mainly store the shared models and service objects between different rails apps (let's say, named `project_frontend` and `proejct_backend` )

``` ruby
module ProjectCore
  class Lyric < ActiveRecord::Base
  end
end
```

And then I define a Form object in `project_backend`.
``` ruby
class CreateLyricForm < Reform::Form
  property :data

  model ProjectCore::Lyric  
end
```

And related `LyricController` in `project_backend` app
``` ruby
class LyricController < ApplicationController
  def new
     @form = CreateLyricForm.new( ProjectCore::Lyric.new )
  end
end
```

And a form_for statement for this create form in `project_backend` related view
``` ruby
<%= form_for @form do |f| %>
  ....
<% end %>
```

Oh, BTW, and define resouces route for this LyricController as well
``` ruby
Rails.application.routes.draw do
  resources :lyrices
end
```

**This is where the whole story begins**

When the system render the view, it would lookup the registered routes, and return a error:
```
ActionView::Template::Error:
       undefined method `project_core_lyrics_path' for
```
the route_key return by `CreateLyricForm.model_name.route_key` is "proejct_core_lyrics", but it is incorrect in this case. 

After this fix, **intended to correct route_key for namespaced model in Isolated Namepace**, the route_key is corrected to "lyrices".

@apotonick, please let me know what you think about this fix! :-D

### DIFF
--- a/lib/reform/form/active_model.rb
+++ b/lib/reform/form/active_model.rb
@@ -92,7 +92,17 @@ module Reform::Form::ActiveModel
   private
     def active_model_name_for(string)
       return ::ActiveModel::Name.new(OpenStruct.new(:name => string)) if Reform.rails3_0?
-      ::ActiveModel::Name.new(self, nil, string)
+      ::ActiveModel::Name.new(self, determine_namespace(string) , string)
+    end
+
+    def determine_namespace(string)
+      begin
+        string.constantize.parents.detect do |n|
+          n.respond_to?(:use_relative_model_naming?) && n.use_relative_model_naming?
+        end
+      rescue NameError
+        nil
+      end
     end
   end
 end
--- a/lib/reform/form/active_model.rb
+++ b/lib/reform/form/active_model.rb
@@ -96,12 +96,9 @@ module Reform::Form::ActiveModel
     end
 
     def determine_namespace(string)
-      begin
-        string.constantize.parents.detect do |n|
-          n.respond_to?(:use_relative_model_naming?) && n.use_relative_model_naming?
-        end
-      rescue NameError
-        nil
+      return nil unless Module.const_defined?(string)
+      string.constantize.parents.detect do |n|
+        n.respond_to?(:use_relative_model_naming?) && n.use_relative_model_naming?
       end
     end
   end
--- a/lib/reform/form/active_model.rb
+++ b/lib/reform/form/active_model.rb
@@ -75,6 +75,11 @@ module Reform::Form::ActiveModel
     #
     #   class CoverSongForm < Reform::Form
     #     model :song
+    #
+    # or we can setup a isolated namespace model ( which defined in isolated rails egine )
+    #
+    #   class CoverSongForm < Reform::Form
+    #     model "api/v1/song", namespace: "api"
     def model(main_model, options={})
       self.model_options = [main_model, options]
     end
@@ -82,24 +87,19 @@ module Reform::Form::ActiveModel
     def model_name
       if model_options
         form_name = model_options.first.to_s.camelize
+        namespace = model_options.last[:namespace].present? ? model_options.last[:namespace].to_s.camelize.constantize : nil
       else
         form_name = name.sub(/(::)?Form$/, "") # Song::Form => "Song"
+        namespace = nil
       end
 
-      active_model_name_for(form_name)
+      active_model_name_for(form_name, namespace)
     end
 
   private
-    def active_model_name_for(string)
+    def active_model_name_for(string, namespace=nil)
       return ::ActiveModel::Name.new(OpenStruct.new(:name => string)) if Reform.rails3_0?
-      ::ActiveModel::Name.new(self, determine_namespace(string) , string)
-    end
-
-    def determine_namespace(string)
-      return nil unless Module.const_defined?(string)
-      string.constantize.parents.detect do |n|
-        n.respond_to?(:use_relative_model_naming?) && n.use_relative_model_naming?
-      end
+      ::ActiveModel::Name.new(self, namespace, string)
     end
   end
 end
--- a/test/active_model_test.rb
+++ b/test/active_model_test.rb
@@ -19,6 +19,7 @@ class NewActiveModelTest < MiniTest::Spec # TODO: move to test/rails/
   describe "::model_name" do
     it { form.class.model_name.must_be_kind_of ActiveModel::Name }
     it { form.class.model_name.to_s.must_equal "NewActiveModelTest::Song" }
+    it { form.class.model_name.route_key.must_equal "new_active_model_test_songs" }
 
     let (:class_with_model) {
       Class.new(Reform::Form) do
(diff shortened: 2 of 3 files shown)


lib/reform/form/active_model.rb
```diff
@@ -92,7 +92,17 @@ def model_name
...
-      ::ActiveModel::Name.new(self, nil, string)
+      ::ActiveModel::Name.new(self, determine_namespace(string) , string)
+    end
+
+    def determine_namespace(string)
+      begin
+        string.constantize.parents.detect do |n|
+          n.respond_to?(:use_relative_model_naming?) && n.use_relative_model_naming?
+        end
+      rescue NameError
```
apotonick: Cool! But why the `rescue` here? Can't we do that without the `begin`/`rescue`?

simoniong: This is a safe guarantee incase the string passed-in can not be constantize. Any good solution for this?

simoniong:
OH, there's couple of errors for missing route_key in Rails 3.0, I will see if I can fix it tonight or tomorrow!

lib/reform/form/active_model.rb
simoniong: OH, just found a nice method defined in ActiveSupport [here](http://apidock.com/rails/ActiveSupport/Inflector/safe_constantize), but it can only support rails version >= 3.2.1... Still no help, may be we can wrapped this kind of logic into a helper,@apotonick, what do you think?

simoniong:
@apotonick sorry, kind of busy last few days, I found that isolated rails engine is only available for rails >= 3.1, so I would suggest to skip those failed test cases if under rails 3.0, here's some example code:
```ruby
unless Reform.rails3_0?
  it { class_with_model.model_name.route_key.must_equal "albums" }
end
```

What do you think?

apotonick:
Yeah, I had a feeling that this feature was added recently. I am happy to run your tests in a condition, please add it to the PR. ❤️

simoniong:
@apotonick , just added filter codes that make test case pass! Please let me know if we need more refactor work!

lib/reform/form/active_model.rb
apotonick: What about removing that `begin` block and using that other method? I hate catching exceptions, I've seen that done a lot in Rails core and it burned people!!!

fernandes:
```ruby
def determine_namespace(string)
  return nil unless Module.const_defined?(string.capitalize)
  string.constantize.parents.detect do |n|
    n.respond_to?(:use_relative_model_naming?) && n.use_relative_model_naming?
  end
end
```

should work ;)

simoniong:
Hi, @apotonick , seems `Module.const_defined?` behave differently in 1.9.3 and 2.0 compared with 2.1 & 2.2...I need some time to investigate with it.

simoniong:
for example, when we define a module like this:
```
module A
  module B
     C = 1
  end
end
```

`Module.const_defined?("A::B::C")` would return `true` for ruby version >=2.1
`Module.const_defined?("A::B::C")` would raise error when ruby version< 2.1

I have a solution, but it seems take more code and logics, for example to have a helper method to make this work, but still not elegant.
``` ruby
    def defined_klass?( string )
      sum = Module
      defined = true
      string.split("::").each do |klass|
        if sum.const_defined?(klass)
          sum = sum.const_get(klass)
        else
          defined = false
          break
        end
      end
      defined
    end
```

@apotonick , any better solution?

fernandes:
yeah @simoniong,

ruby < 2.1 don't check full strings, my bad... you need to check each children like you did, I see no better solution than yours. good point

ps: Could put in less lines of code, but would compromise readability, so I'm +1 for yours

apotonick:
I would prefer a less magic approach. What if we allow forms to define their namespace declaratively?

```ruby
class Song::Form < Reform::Form
  model Namespace::Song, namespace: ""
```

This is way simpler to implement and we don't need this `const_defined?` and all that - this will always guess the wrong and we will be bombarded with bug reports.

Let's go the explicit Trailblazer way!

apotonick:
Or what about allowing strings in `::model`?

```ruby
model "song" # will set model_name to Song.
```

fernandes:
guess the wrong and we will be bombarded with bug reports.

Are you talking about rails autoloading? :stuck_out_tongue_winking_eye:

simoniong:
I would prefer to go for a easier but clean way ( But this might be a breaking change, right? :P ). @apotonick , can you give me more example, so that I can fully understand what your intended is. 

@fernandes  oh, I thought that comment is from @apotonick , sorry, my bad, man! thanks for your help and idea :+1: let's make it rock and roll to chasing elegant!

apotonick:
Haha, this is a fully-compatible extension to what we already have. People can simply turn off/extend/change their `model_name`. For Reform, it doesn't really matter what the value is: we simply convert it to an `AM::Name` in `Form::model_name`, similar to what you did already. https://github.com/apotonick/reform/blob/master/lib/reform/form/active_model.rb#L82

So, let's say I'd do 

```ruby
model "api/v1/song", this will result in a `AM::Name` with `Api::V1::Song` and Rails can do whatever it wants with that.

The whole namespacing in Rails currently is very dodgy and the less guessing we introduce the better.

**Comment by fernandes** [comment]
@simoniong no problem ;)

**Comment by simoniong** [comment]
So, let me guess what your intention is :)

in your case, this is non isolated model version:
``` ruby
  model "api/v1/song"
```
This would result with `Api::V1::Song`, and instance of AM::Name as a result for `model_name`
``` ruby
ActiveModel::Name.new(self, nil, `Api::V1::Song` )
```
oh, **btw user don't need to explicit the namespace.**

And in a isolated model version( for example, I have a rails engine call `Api`):
``` ruby
  model "api/v1/song", namespace: "api"
```
This would result with the same string `Api::V1::Song`, and instance of AM::Name as a result for `model_name`

```ruby
ActiveModel::Name.new(self, 'Api', `Api::V1::Song` )
```

With this kind of spec, we can simple remove the `const_defined?` and all the shit away!

Is this correct?

simoniong:
Hi, @apotonick , any comment?

apotonick:
Sorry for the delay brother!

I don't really understand why `ActiveModel::Name` needs to know the namespace - can you elaborate on that, please?

However, yes, you're right - if we need that namespace option for the AM stuff, then this option `:namespace` makes sense. :shipit:

gdott9:
Hello,

`ActiveModel::Name` has to know the namespace when inside an isolated namespace (http://api.rubyonrails.org/classes/Rails/Engine.html#method-c-isolate_namespace) because the route keys are not prefixed by the namespace in this case.

`ActiveModel::Name` with namespace
```ruby
ActiveModel::Name.new(self, Admin, 'Admin::User')
=> #<ActiveModel::Name:0x007f3e0b9dc2b8
...
 @name="Admin::User",
 @param_key="user",
 @plural="admin_users",
 @route_key="users",
 @singular="admin_user",
 @singular_route_key="user",
 @unnamespaced="User">
```
`ActiveModel::Name` without namespace
```ruby
ActiveModel::Name.new(self, nil, 'Admin::User')
=> #<ActiveModel::Name:0x007f3e0b970c20
...
 @name="Admin::User",
 @param_key="admin_user",
 @plural="admin_users",
 @route_key="admin_users",
 @singular="admin_user",
 @singular_route_key="admin_user">
```
So, if I use `model :'admin/user'` in my form, the name is correct but not the route keys and if I use `model :user`, the route keys are good but not the model name.

The namespace option is simple and seems to solve this problem in a less magical way.

apotonick:
Thanks @gdott9 for explaining. To me, this sounds like a design flaw in Rails itself. The way routes are addressed is wrong.

However, this is not our problem: we simply allow specifying `:namespace` and it's all good. Ready for it, @simoniong ? :laughing: :rocket:

simoniong:
@apotonick  @gdott9 ,  Sorry for the late reply, kind of busy holiday for Lunar New Year last few weeks. :beers:
I am ok with it, and the code should be ready! :pizza: :beers: :kissing:

gdott9:
Hello @apotonick 

Any news about this ? Could you merge the pull request ?
I rebased it on master and put everything into one commit if you prefer https://github.com/gdott9/reform/commits/fix_isolated_engine_namespace.

apotonick:
Awesome, thanks! Will merge tomorrow!

fran-worley:
@simoniong I'm sorry this got missed. ActiveModel related code has moved to reform-rails gem and I've just looked and the namespace option is supported already. https://github.com/trailblazer/reform-rails/blob/master/lib/reform/form/active_model.rb#L44-L55

I'm closing this for now, you're welcome to raise another PR in reform-rails if we've not covered it sufficiently.

PR #193, opened 2015-01-23
Memoize `#model_name` in `Reform::Form::ActiveModel`

gdott9:
This PR stores the form model name in an instance variable to only create one `ActiveModel::Name` object.

### DIFF
--- a/lib/reform/form/active_model.rb
+++ b/lib/reform/form/active_model.rb
@@ -80,13 +80,15 @@ module Reform::Form::ActiveModel
     end
 
     def model_name
-      if model_options
-        form_name = model_options.first.to_s.camelize
-      else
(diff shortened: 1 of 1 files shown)


apotonick:
Thanks! Not a fan of keeping internal state, though. :surfer:  What's your objective with that PR?

Memoizing is a pattern found in Rails and an indicator for a wrong API: if you need to compute a value several times, your public API is leaky and the way you pass arguments needs to be optimised.

I see your point, though, as this is cached in production.

gdott9:
I was working on the bug with isolated namespaces and noticed that `#model_name` is called several times when generating a form with Formtastic.
No problem, I see your point too and maybe I should rather look into Formtastic why it needs to call this method several times.

Thank you for your answer (and your awesome gems). :smiley:

PR #194, opened 2015-01-23
Add support for isolated namespaces in `#model_name`

gdott9:
When the class is in an isolated namespace, the namespace has to be
defined to generate the correct route key in `ActiveModel::Name`.


gdott9:
Oops, I didn't see a PR was already open for this issue.

apotonick:
:grin:

PR #196, opened 2015-01-26
Add support for DateTime select in `MultiParameterAttributes`

gdott9:
Rails has an helper for a datetime_select (http://api.rubyonrails.org/classes/ActionView/Helpers/DateHelper.html#method-i-datetime_select) which cannot be used with Reform because the `DateParamsFilter` will try to convert it into a date without using the params for hours (4i) and minutes (5i).

I tried to handle those params too and to generate a `Time` object using the correct timezone.

I also added a `#dup` on the `params` hash. I don't know what you think about that, but at first I didn't understand why the `params` hash in my controller was not the same as the parameters I was sending in my form so I think it's better to work with a copy of this hash.

### DIFF
--- a/lib/reform/form/multi_parameter_attributes.rb
+++ b/lib/reform/form/multi_parameter_attributes.rb
@@ -4,41 +4,47 @@ Reform::Form.class_eval do
       base.send(:register_feature, self)
     end
 
-    class DateParamsFilter
+    class DateTimeParamsFilter
       def call(params)
-        date_attributes = {}
-
-        params.each do |attribute, value|
+        params.each_with_object({}) do |(attribute, value), hsh|
           if value.is_a?(Hash)
-            call(value) # TODO: #validate should only handle local form params.
+            hsh[attribute] = call(value) # TODO: #validate should only handle local form params.
           elsif matches = attribute.match(/^(\w+)\(.i\)$/)
             date_attribute = matches[1]
-            date_attributes[date_attribute] = params_to_date(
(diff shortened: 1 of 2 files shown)


apotonick:
Thanks! I merged your stuff here: https://github.com/apotonick/reform/compare/datetime

However, I had to fix the `dup`ing of `params` to be recursive. Yours just ran it on the first level.

What's the deal with this, anyway? If we just changed `params.delete("release_date(1i)")` to simply read from the hash, wouldn't that be better? So we can avoid that duplication stuff? Do we really need to `delete` the multi-params from `params`?

gdott9:
If we don't delete the multiparameter attributes from the hash, `#validate` will receive a hash with multiparameter attributes and simple attribute for the same field. I don't know how these conflicting attributes are handled in Rails and if it will always be processed the same way so I think it's better to clean the hash to be sure.

I pushed a new commit where I replaced the duplication and the delete with the creation of a new hash. I don't know which one you prefer.

apotonick:
That's alright, I'll leave the one we have now. Thanks!!!

kirs:
Thanks for the feature @gdott9! I'm using your fork in my project now.

kirs:
Looking forward for the new release of reform including this feature @apotonick

apotonick:
I love releasing gems in the early mornings. https://github.com/apotonick/reform/commit/79f1e45d6aebdf1aef67383a8b1a0eea3bfcb324 :heart:

PR #197, opened 2015-01-30
Add Section to README on Virtus & Arrays of IDs

posgarou:
I think this is a not uncommon situation, and when I ran into it it took a little bit of research to find a good solution.  (I wanted to avoid the custom `schools=` method, since that seemed like a "last resort" solution.)

I think this technique has a lot of benefits, especially readability and the ability to easily add min/max testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -624,6 +624,37 @@ Coercion only happens in `#validate`.
 form.written_at #=> <DateTime "2014 September 26 00:00">
 ```
 
+### Using Virtus to Coerce an Array of IDs into Objects/Models
+
+You may at times need to coerce an array of model IDs into their corresponding models.
+
+For example, imagine a Student has_and_belongs_to_many Schools.  You may have a select box or similar form element on a StudentForm that includes a number of School id/name option pairings.  Given this setup, when the form is submitted the response params would include a list of ids, e.g. `"schools" => ["1", "3", "5", "7"]`.
+
+In this situation it is easiest to make a custom Virtus::Attribute to use for your coercion, as in the following example:
+
+```ruby
+class StudentForm < Reform::Form
+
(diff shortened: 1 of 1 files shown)


apotonick:
Hey Ryan, thanks!

Can you document this in a Wiki page? This is very specific information that might blow up our README.

Also, why don't you use `:prepopulate_if_empty` or `:populator` for that? That option's intention is exactly that: process the incoming values and transform them into models. Have you considered one of those options?

Anyway, cool idea that you came up with - I never thought of that haha. :+1:

posgarou:
Sure, I can put that section in the Wiki.

Now, please correct me if I'm misunderstanding the feature, but prepopulating is used for setting up the form object for display but has no role in processing the received data after the data is passed back.

In this case (not uncommon, I think) the select field contains id/name pairings for already existing objects, and this data is loaded in via AJAX.

The problem I was running into was that I had to write a custom setter method to convert the array of ids back into an array of models to store.  (And this conversion or something like it had to happen since it was a HABTM.)

Using the Virtus technique above obviates the need for writing that custom setter method in (I think) a much more readable way.

That said, I haven't played around with prepopulating a lot, so please correct me if my understanding of the feature is off.

posgarou:
Also, as an aside, thanks for your work on Reform.  It's made my code so much more readable and maintainable since I started using it.  :+1:

apotonick:
@posgarou Haha, I am talking about population, not about pre-population. This is discussed in [my book in chapter 5](https://leanpub.com/trailblazer) and should clarify your misunderstandings. Please please please do let me know if it doesn't!

Thanks for your support, go spread the word!

posgarou:
Good to know!  Thanks!

PR #198, opened 2015-02-07
Correct Failing Tests

posgarou:
(1) First test had extra 's'.
(2) Second test incorrectly tested instance_of rather than kind_of when testing the reflection.  (See the [class hierarchy in active_record/reflection.rb](https://github.com/rails/docrails/blob/master/activerecord/lib/active_record/reflection.rb#L165-L174).)

### DIFF
--- a/test/validate_test.rb
+++ b/test/validate_test.rb
@@ -261,7 +261,7 @@ class ValidateTest < BaseTest
       before { subject.validate({}).must_equal false }
 
       it { subject.errors.messages.must_equal(
-        :songs => ["is too short (minimum is 1 characters)"],
+        :songs => ["is too short (minimum is 1 character)"],
         :hit   => ["can't be blank"]) }
     end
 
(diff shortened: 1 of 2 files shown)


posgarou:
These tests were failing using Rails 4.2.0, by the way.  I'm guessing they passed in earlier versions of Rails.

test/validate_test.rb
```diff
@@ -260,9 +260,10 @@ class ValidateTest < BaseTest
     describe "invalid" do
       before { subject.validate({}).must_equal false }
 
-      it { subject.errors.messages.must_equal(
-        :songs => ["is too short (minimum is 1 characters)"],
-        :hit   => ["can't be blank"]) }
+      it do
+        subject.errors.messages[:hit].must_equal(["can't be blank"])
```
apotonick: We should add a test that no other keys are in `messages` (for completeness). Thanks for your work, can you please add this? :beers:

posgarou:
@apotonick With this done, maybe it makes sense to add Rails 4.2 and Ruby 2.2.1 to `.travis.yml`?

The tests all pass locally running those two versions, but I didn't run all the permutations.

apotonick:
Please add it to `.travis.yml`! :heart:

PR #199, opened 2015-02-07
Make namespaced forms return model for #to_model.

posgarou:
While working with this basic setup today

```ruby
class PostForm < Reform::Form

  class Admin < PostForm; end

  class Normal < PostForm; end

  # def to_model
  #   model
  # end
end
```

I discovered that when sending my view a PostForm::Normal instance, the path generated for the form action (`= form_for @form do |post_form|`) was defaulting to `post_form_normals_path`.  After some digging, I found that this was due to `#to_model` and that uncommenting the above-defined `#to_model` method gave the expected behavior.

This change implements the expected behavior for non-composed forms and preserves the existing behavior for composed forms.

### DIFF
--- a/lib/reform/form/active_model.rb
+++ b/lib/reform/form/active_model.rb
@@ -54,8 +54,11 @@ module Reform::Form::ActiveModel
 
       delegates :model, *[:persisted?, :to_key, :to_param, :id] # Uber::Delegates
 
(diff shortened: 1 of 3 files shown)


fran-worley:
@posgarou all rails/ ActiveModel related code has been moved to the reform-rails gem. I'm going to close this PR, but feel free to reopen in reform-rails if you want. 

Thanks!

PR #212, opened 2015-03-19
Expose contract fields

fernandes:
WIP: Need to check the [Formtastic PR](https://github.com/justinfrench/formtastic/pull/1142), then I add tests and doc before merge, just opened this PR so we can discuss seeing some code.

This PR is related to https://github.com/justinfrench/formtastic/issues/1141

My proposal to solve the problem on the issue above, without affecting the current Formtastic behavior with AR models.

When a Reform object is passed to Formtastic builder, its InputsHelper tries to discover the object field, if the object respond to `form_object_fields` (as suggested in this PR), Formtastic knows how to build the form for the object based on the specified fields.

//cc @justinfrench

### DIFF
--- a/lib/reform/contract.rb
+++ b/lib/reform/contract.rb
@@ -26,6 +26,10 @@ module Reform
 
     RESERVED_METHODS = [:model, :aliased_model, :fields, :mapper] # TODO: refactor that so we don't need that.
 
+    # This method is used so a form builder (specifically Formtastic) can infer the input fields
+    def form_object_fields
+      fields.to_h.keys
+    end
 
     module PropertyMethods
       def property(name, options={}, &block)


apotonick:
Hi @justinfrench - where does formtastic add `#form_object_fields` to `ActiveModel`? Is that a method you guys add as your formtastic API?

@fernandes We can add this for compatibility with formtastic, but this has to go into a separate module and has to be included explicitely to document API dependencies. We do the same with simple_form and ActiveModel. Thanks!!! :heart:

justinfrench:
@apotonick answered [here](https://github.com/justinfrench/formtastic/pull/1142) I think.

fran-worley:
I'm closing this as it needs still some changes prior to merging see @apotonick comment 3 Apr 2015 and it's been a while since the last update. 

If you still want to provide support for this you will need to PR the reform-rails gem as model_reflections have been moved. 

Thanks!

fernandes:
ok @fran-worley , I'm not using ActiveAdmin anymore, no problem to close. thank you!

PR #221, opened 2015-04-09
Virtual: true doesn't prevent from writing to object

freemanoid:
Virtual attributes don't really virtual. Reform returns them in `save` block. 
Adding `writeable: false` helps. That's weird because `virtual: true` [sets both `writeable` and `readable` to false][1].
Here is a test to reproduce.

[1]: https://github.com/apotonick/reform/blob/master/lib/reform/contract.rb#L43-L44

### DIFF
--- a/test/save_test.rb
+++ b/test/save_test.rb
@@ -103,6 +103,28 @@ class SaveTest < BaseTest
     it { label.saved?.must_equal nil }
   end
 
+  describe "virtual: true" do
+    Reform::Form.reform_2_0!
+
+    let(:form) do
+      Class.new(Reform::Form) do
+        property :some_unexisted_attribute, virtual: true
+
+        # Unexpectedly adding "writealbe: false" helps (despite the reform_2_0!)
+        # property :some_unexisted_attribute, virtual: true, writeable: false
+      end
+    end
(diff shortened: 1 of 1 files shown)


apotonick:
This is because to activate the new `:virtual` behaviour, you need to call `reform_2_0!`: https://github.com/apotonick/reform/wiki/Migration-Guide#migration-guide-for-new-20-semantics-optionally-available-since-12

freemanoid:
I call it in the test https://github.com/apotonick/reform/pull/221/files#diff-833e6ed7dedaccdacdd19698f62fdda1R107

apotonick:
Aaaah, sorry, now I got it: According to you, the virtual attribute should not be in the nested hash when calling `#save` with a block? I decided that I want this in this hash, that's why it's there.

Reason for this is that you might need those values when saving or further processing the form, the only clean way to retrieve this is via the nested hash.

Maybe something like this?

```ruby
form.save(include_virtual: false) do .. end
```

freemanoid:
At first it's a bit inconsistent because adding `writeable: false` removes these properties from the hash. 
At second I think that if the property is virtual you should not expect to get it in the hash. If you want to use it, you should make it `writeable`.

PR #222, opened 2015-04-15
Include rails 4.2 in .travis.yml

posgarou:
You mentioned in #198 that you'd like `.travis.yml` upgraded.  I saw that #209 had already upgraded the Ruby versions, but here it is with Rails 4.2 added.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ rvm:
   - 2.0.0
   - 1.9.3
 gemfile:
+  - gemfiles/Gemfile.rails-4.2
   - gemfiles/Gemfile.rails-4.0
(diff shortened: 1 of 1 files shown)


posgarou:
It looks like the Rails 4.2 tests failed with `cannot load such file -- uber/inheritable_attr (LoadError)`.

(1) Any idea why?
(2) Do you want me to list it as an allowable failure for now or concentrate on making the tests pass?

apotonick:
Thanks! Where _is_ this 4.2 Gemfile? It looks as if the uber gem is not required properly?!

posgarou:
That would be the problem.  :)  I'll fix it sometime this week and see if
everything works.

PR #224, opened 2015-05-22
Compatibility for Mongoid

arturictus:


Titinux:
:+1:

I will test that as soon as possible.

apotonick:
WOW! :heart: :heart_eyes: :guitar: :rocket:

Titinux:
It works great :kissing:

arturictus:
Hi @Titinux and @apotonick , 
I saw you have some TODOs in the  _lib/reform/form/active_record.rb_ and I copied the file to _lib/reform/form/active_record.rb_ and just changed to Mongoid.
I think it needs a refactor so we do not duplicate some of this code.
I'm not sure how it has to be done.
Do we create a module at _lib/reform/form/orm.rb_ for the common logic and we  do the TODOs?
Any Idea?

Titinux:
It would be great. We should put at least model_for_property method in common. The orm module seems good but let's hear @apotonick about this.

arturictus:
:+1: :smile:

apotonick:
Loving the idea of `lib/reform/form/orm.rb` even though I had to read twice to see you meant to say "orm". Do it!!!!!!!!! :heart:

arturictus:
Hi guys,
I've done the ORM module, check it out if is fine for you.
:rocket: :smile:

lib/reform/form/orm.rb
```diff
@@ -0,0 +1,29 @@
+module Reform::Form::ORM
+  module Base
```
apotonick: Looks great, @arturictus ! One thing: I don't use `Base` as a name anymore, that's as redundant as `Manager` or `Object`. Just call it `ORM`. `Base` is one of the many things we should _not_ copy from Rails.

arturictus:
Hi @apotonick , I find It redundant too, but I need to namespaces, one for the uniqueness validator and another one forma the adapter It self. Do you think in a better way?. I don't want to include the uniqueness validator in the adapter.

lib/reform/form/mongoid.rb
```diff
@@ -0,0 +1,39 @@
...
+    else
+      'Validations'
+    end
+  end
+
+  def self.mongoid_is_4_or_more?
+    Mongoid::VERSION.split('.').first >= '4'
+  end
+
+  UniquenessValidator = Class.new("::Mongoid::#{mongoid_namespace}::UniquenessValidator".constantize) do
```
apotonick: What's `mongoid_namespace`? This kind of stuff is really really dangerous and it will break for people and then they will blame me.

apotonick: Ah sorry, ok. Will merge and do some restructuring so you see what I mean (better than to discuss that over and over, right?).

arturictus: ok :)

2015-05-26 0:13 GMT+02:00 Nick Sutterer <notifications@github.com>:

> In lib/reform/form/mongoid.rb
> <https://github.com/apotonick/reform/pull/224#discussion_r30996235>:
>
> > +  end
> > +
> > +
> > +  def self.mongoid_namespace
> > +    if mongoid_is_4_or_more?
> > +      'Validatable'
> > +    else
> > +      'Validations'
> > +    end
> > +  end
> > +
> > +  def self.mongoid_is_4_or_more?
> > +    Mongoid::VERSION.split('.').first >= '4'
> > +  end
> > +
> > +  UniquenessValidator = Class.new("::Mongoid::#{mongoid_namespace}::UniquenessValidator".constantize) do
>
> Ah sorry, ok. Will merge and do some restructuring so you see what I mean
> (better than to discuss that over and over, right?).
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/apotonick/reform/pull/224/files#r30996235>.
>



-- 
Artur Pañach

apotonick:
@arturictus I merged your changes here: https://github.com/apotonick/reform/commit/2faf2f8f4c60c53e37942425ded3f8972d86561c

I decided to _not_ use ActiveRecord's `UniquenessValidator` anymore as it writes to the model, and will now enforce Reform's `FormUniqueValidator`. I'd love to make that one compatible with Mongoid - do you think we can do it (while supporting `scope`, too?)?

The reason is that I get a lot of issues from AR's writing validator, so best is not to use it at all.

apotonick:
@arturictus You are in master now. Tests for Rails 3.1 and 3.2 are broken, but I guess it's a minor issue you can fix quickly *no pressure*_!!!_ :wink: 

One thing: Forgot what I said above, the uniqueness in AR and Mongoid still gets passed to the model class and messes around with model attributes, etc. Since this is an optional feature, users have to take care about their data integrity themselves (or make Rails core fix that problem in AM::Validations).

Our alternative implementation `unique: true` won't work with Mongoid, _yet_. Can you have a look how that could work? Here's the test where you could start. https://github.com/apotonick/reform/blob/master/test/unique_test.rb

I documented all that here: http://trailblazerb.org/gems/reform/validation.html

arturictus:
Hi Nick,
Great!! :).
I'll take a look to the 'unique: true' attribute and the Rails 3.1 and 3.2
compatibility.
I'm quite busy right now but I'll do it as soon as I can.
Thanks


2015-07-02 3:43 GMT+02:00 Nick Sutterer <notifications@github.com>:

> @arturictus <https://github.com/arturictus> You are in master now. Tests
> for Rails 3.1 and 3.2 are broken, but I guess it's a minor issue you can
> fix quickly *no pressure**!!!* [image: :wink:]
>
> One thing: Forgot what I said above, the uniqueness in AR and Mongoid
> still gets passed to the model class and messes around with model
> attributes, etc. Since this is an optional feature, users have to take care
> about their data integrity themselves (or make Rails core fix that problem
> in AM::Validations).
>
> Our alternative implementation unique: true won't work with Mongoid, *yet*.
> Can you have a look how that could work? Here's the test where you could
> start. https://github.com/apotonick/reform/blob/master/test/unique_test.rb
>
> I documented all that here:
> http://trailblazerb.org/gems/reform/validation.html
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/apotonick/reform/pull/224#issuecomment-117870725>.
>



-- 
Artur Pañach

PR #233, opened 2015-06-15
Improve unique validator on model update

fernandes:
Unique validator was searching if there was some model with value the same as the form, but when we are updating (and obviously the object is already persisted), the sql will find an object, the same as we are trying to update, so we need to consider two cases:

1. on create:

search for any object with field with same value (case works perfectly)

2. on update:

search for any object with field with same value AND id different the form's model id, so we will never find the same object (now this PR is considering this case!)

and now unique validation works perfectly

### DIFF
--- a/lib/reform/form/validation/unique_validator.rb
+++ b/lib/reform/form/validation/unique_validator.rb
@@ -1,14 +1,13 @@
 class Reform::Form::UniqueValidator < ActiveModel::EachValidator
   def validate_each(form, attribute, value)
-    if form.model.persisted?
-      if form.model.class.where("#{attribute} = ? AND id <> ?", value, form.model.id).size > 0
-        form.errors.add attribute, "#{attribute} must be unique."
-      end
-    else
-      if form.model.class.where("#{attribute} = ?", value).size > 0
-        form.errors.add attribute, "#{attribute} must be unique."
-      end
-    end
+    # search for models with attribute equals to form field value
+    query = form.model.class.where(attribute => value)
+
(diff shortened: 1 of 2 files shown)


lib/reform/form/validation/unique_validator.rb
```diff
@@ -1,7 +1,14 @@
 class Reform::Form::UniqueValidator < ActiveModel::EachValidator
   def validate_each(form, attribute, value)
-    if form.model.class.where("#{attribute} = ?", value).size > 0
-      form.errors.add attribute, "#{attribute} must be unique."
+    if form.model.persisted?
+      if form.model.class.where("#{attribute} = ? AND id <> ?", value, form.model.id).size > 0
```
freemanoid: I think it's better to not interpolate an `attribute` in sql query like this because of possible escaping and quoting problems. So this can be changed to 
```ruby
query = form.model.class.where(attribute => value)
if form.model.persisted?
  query = query.merge(form.model.class.where("id <> ?", form.model.id)
end

if query.any?
    form.errors.add(attribute, "#{attribute} must be unique.")
end
```

What do you think?

fernandes: hey @freemanoid, I just fixed what was coded before, but what you said makes sense... I'm gonna update PR

thank you

fernandes: refactored :smile:

apotonick:
Merged in reform-2. :kiss:

fernandes:
:kissing_cat:

Issue #244, opened 2015-07-13
Model name in ActiveModel validations for custom error message translations is always "reform/form"


tilod:
Unfortunatly I found another bug. When using the standard error translations everything works (because they are not in a model namespace), but when using custom translations, I18n uses a wrong model name for the lookup, namely "reform/form".

I put another PR together with tests for this. In short:

```yaml
en:
  # custom validation error messages
  activemodel:
    errors:
      models:
        song:
          attributes:
            title:
              custom_error: Custom Error Message
```

```ruby
class SongForm < Reform::Form
  model :song
  property :title

  validate do
    errors.add :title, :custom_error_message
  end
end
```

Triggering the validation will add `"translation missing: en.activemodel.errors.models.reform/form.attributes.title.custom_error_message"` to the errors hash. Notice the model name it uses for the translastion lookup: Although I used `model :song` to set it to "song", it's "reform/form".

The weird thing is, `form.class.model_name.human` (which should be called according to the Rails internationalization guide) actually returns "Song". I also tried overriding `::model_name` in the form which lead to the same result: calling `form.class.model_name` returned the correct result, but it's not used in the translation.

apotonick:
Ah yeah I remember that from years ago.

Another reason I will push things towards a clean implementation using Lotus::Validations. ActiveModel is all but a standard and makes _so many assumptions_ about your "model" that it's impossible to keep track if you're "compliant" to the "standard", even though there's wannabe lint tests available (outdated, of course).

Man, thanks so much for the test, they really really help!

tilod:
You're welcome.

I played a little bit and found out, the reason for the bug is the implementation of `::model_name` in `Reform::Form::ActiveModel::Validations::Validator`. It is hardcoded to return `Reform::Form`. It should actually return what the model in the form returns (and maybe `Reform::Form` as fallback). Given it's a class method which should delegate to another class method, which itself may later (after the `included` hook is run) be overridden (by calling `::model` in the form)… oh well, you will have fun with this one, I guess.

spectator:
I run into similar issue but with attribute translations.

```yml
en:
  activerecord:
    attributes:
      albums/album:
        song_ids: "Songs"
```

this used to work, but with Reform 2.0.1 I get "Song ids" which is fallback behavior for `human_attribute_name`

apotonick:
[...]

PR #246, opened 2015-07-13
add tests for #244

tilod:
Here are the (failing) tests for #244.


apotonick:
Thanks, merged, but for some reasons GH doesn't "see" it as merged.

PR #262, opened 2015-07-20
Acceptance validator

spectator:
Reform 2.0.1, Rails 4.1.10

I was trying to upgrade Trailblazer and friends to latest version and run into a couple of issues, this is one of them - acceptance validation no longer works.

### DIFF
--- a/lib/reform/form/validation/accept_validator.rb
+++ b/lib/reform/form/validation/accept_validator.rb
@@ -1,0 +1,22 @@
+# Rails acceptance validation that does not write to model.
+class Reform::Form::AcceptValidator < ActiveModel::EachValidator
+  def initialize(options)
+    super({ accepted: ["1", true] }.merge!(options))
+  end
+
+  def validate_each(form, attribute, value)
+    unless accepted_option?(value)
+      form.errors.add(attribute, :accepted, options)
+    end
+  end
+
+  private
+
+  def accepted_option?(value)
+    Array(options[:accepted]).include?(value)
+  end
+end
+
+Reform::Form::ActiveModel::Validations::Validator.class_eval do
+  AcceptValidator = Reform::Form::AcceptValidator
+end
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+## Unrelased
+
+* Added `validates accept: true` validation as a replacement for `validates acceptance: true`. This validation will not write anything to model. Validation implementation is courtesy of Rails team, thanks!
+
 ## 2.0.3
 
 * `Form#valid?` is private now. Sorry for the inconvenience, but this has never been documented as public. Reason is that the only entry point for validation is `#validate` to give the form as less public API as possible and minimize misunderstandings.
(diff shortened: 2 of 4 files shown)


apotonick:
We have to write our own acceptance validator. The Rails one does wildly assume shit, writes to the model and I am not willing to fix Rails code anymore. Please either use Lotus validations or add a fix like this: https://github.com/apotonick/reform/blob/master/lib/reform/form/validation/unique_validator.rb

spectator:
@apotonick Force pushed acceptance validator implementation. Not very happy with name `accept: true`, will be happy to change to something else if you have better ideas.

lib/reform/form/validation/accept_validator.rb
```diff
@@ -0,0 +1,22 @@
+# Rails acceptance validation that does not write to model.
+class Reform::Form::AcceptValidator < ActiveModel::EachValidator
+  def initialize(options)
+    super({ accepted: ["1", true] }.merge!(options))
+  end
+
+  def validate_each(form, attribute, value)
+    unless accepted_option?(value)
+      form.errors.add(attribute, :accepted, options)
```
spectator: I wasn't sure what to use here as a message. In `UniqueValidator` this message is hardcoded, but we still inherit from `ActiveModel::EachValidator` so as for me it makes more sense to use I18n key.

Let me know if you want "key version" or "hardcoded message". I can also update `UniqueValidator` if you decide in favor of "key version".

apotonick:
We don't have to worry too much about details here, as this is on the declining ActiveModel line which we will soon drop. I'm happy to merge, let me check it out. Can't we "override" the `acceptance: true` switch from AM?

spectator:
I thought about that but it looks like that switch is dynamic based on validator class name and the way to override it would be overriding the whole validator which would break stuff that doesn't use trailblazer yet :(

joshuapaling:
For anyone else reading this, as an interim solution, what the validator @spectator created in an initializer:

`config/initializers/reform.rb`
```
# https://github.com/apotonick/reform/pull/262
# acceptance validator no longer works in reform 2,
# so this is a workaround, until the PR linked to above gets merged.
class Reform::Form::AcceptValidator < ActiveModel::EachValidator
  def initialize(options)
    super({ accepted: ["1", true] }.merge!(options))
  end

  def validate_each(form, attribute, value)
    unless accepted_option?(value)
      form.errors.add(attribute, :accepted, options)
    end
  end

  private

  def accepted_option?(value)
    Array(options[:accepted]).include?(value)
  end
end

Reform::Form::ActiveModel::Validations::Validator.class_eval do
  AcceptValidator = Reform::Form::AcceptValidator
end
```

and then you're good to go using something like this in your form object: 

```
validates :agreed_to_terms, accept: { message: 'Please agree to the terms' }
```

smathy:
+1 on merging this please.

apotonick:
I am thinking about a reform-activemodel gem where we can put this kind of code and the existing AM-related implementation in Reform. 

I really want to get rid of ActiveModel and drop support (by this I mean that _I_ don't want to work on it but you're free to do), and put the reform-activemodel gem into the hands of the Trailblazer community.

What do you think about this?

smathy:
Definitely sounds like a good direction to me (for what that's worth), the less coupling with Rails (or anything) in the core gems the better IMHO.

spectator:
Good idea.

apotonick:
This is fixed via #310. Thanks all, especially @monkbroc.

Issue #241, opened 2015-07-13
Calling #valid? on form object adds duplicate messages to errors array


npearson72:
```
=> {:name=>["can't be blank", "can't be blank", "can't be blank"],
 :quantity=>
  ["must be greater than or equal to 1000",
   "must be greater than or equal to 1000",
   "must be greater than or equal to 1000"]}
```
Each time I call form#valid? reform pushes a duplicate into the array. Is this by design for some reason?

apotonick:
Yes! `#valid?` is not a documented public method. You should use a contract and `#validate` or validate with an empty hash, but the `validate` method isn't designed to be run multiple times, too.

Looking into the code, I see why and should probably fix it :grin: ...

npearson72:
Cool, thanks for the reply. This doesn't happen with Lotus by the way, only with AV.

PR #263, opened 2015-07-20
Failing test to show that `valid?` duplicates messages

spectator:
Reform 2.0.1

I haven't run into this case with normal usage, but I thought it makes sense to report it anyway.

What happens is that calling `valid?` after `validate({})` will duplicate errors message. Given already validated form

```ruby
{:title=>["can't be blank"], :"band.label.name"=>["can't be blank"]}
```

calling `validate?` will do

```ruby
{:title=>["can't be blank", "can't be blank"], :"band.label.name"=>["can't be blank"]}
```

the more times you call `valid?`, the more duplicated messages you get.

### DIFF
--- a/test/errors_test.rb
+++ b/test/errors_test.rb
@@ -142,4 +142,12 @@ class ErrorsTest < MiniTest::Spec
     # to_s is aliased to messages
     it { form.errors.to_s.must_equal "{:\"songs.title\"=>[\"can't be blank\"], :\"band.label.name\"=>[\"can't be blank\"]}" }
   end
+
+  describe "repeated validations" do
+    it do
+      form.validate("title" => "")
+      form.valid?
+      form.errors.messages.must_equal({:title=>["can't be blank"], :"band.label.name"=>["can't be blank"]})
+    end
+  end
 end


spectator:
Ah, this is already reported in #241

apotonick:
`#valid?` is not public API, please don't use it. Reform tries to be semi-functional which means you have to use `#validate` when you want to validate. #thisaintrails

spectator:
This isn't rails for sure, but even Lotus has `valid?` which is and will always be confusing. Any way to make `valid?` private at least?

apotonick:
Yeah we should make it private (I thought it was, already). The thing is, both AM and Lotus validations are one level deeper, they validate state (and AM does a lot of messed up shit to achieve that), but `Reform::Form` allows to change application state before validating the object graph.

In case you didn't know, `Reform::Contract` gives you `#validate` without accepting params, exposing the exact same semantics as `#valid?`.

You're right, we should care to make it private.

spectator:
I'll close this PR then.

Thank you for the great stuff you do!

apotonick:
Thank you! :wink: You can open a PR that makes `#valid?` private in Lotus::Validations and ActiveModel::Validations (both modules in Reform). :+1:

Issue #142, opened 2014-09-16
Exception when invalid date


zubin:
Given a reform class with a date input, eg:

    class SomeForm < Reform::Form
      property :starts_on, type: Date
    end 

When a user submits an invalid date (eg 31 Sept), an exception is raised when reform parses the date params. The exception is simply "invalid date" (ie no exception class). I think it's raised by Date.

Here's the relevant backtrace:

    reform (1.0.1) lib/reform/form/multi_parameter_attributes.rb:30:in `new'
    reform (1.0.1) lib/reform/form/multi_parameter_attributes.rb:30:in `params_to_date'

fran-worley:
@zubin any update? I'm going to close this as Multi-parameter attribute handling has moved to reform-rails.

If this is still not working as expected can you create a new issue there?

PR #267, opened 2015-07-21
Catch date parse exception

joealba:
When using multi-parameter attributes for date fields, an invalid date would raise an ArgumentError.  

I added a test for that case.  Then I added a commit to catch that exception and return nil.

params_to_date already has a guard clause that returns nil if year, month or day are blank.  How do you feel about an invalid date behaving the same as that guard clause?

Ref #142

### DIFF
--- a/lib/reform/form/multi_parameter_attributes.rb
+++ b/lib/reform/form/multi_parameter_attributes.rb
@@ -26,12 +26,16 @@ module Reform::Form::MultiParameterAttributes
     def params_to_date(year, month, day, hour, minute)
       return nil if [year, month, day].any?(&:blank?)
 
-      if hour.blank? && minute.blank?
-        Date.new(year.to_i, month.to_i, day.to_i) # TODO: test fails.
-      else
-        args = [year, month, day, hour, minute].map(&:to_i)
-        Time.zone ? Time.zone.local(*args) :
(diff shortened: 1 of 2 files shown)


lib/reform/form/multi_parameter_attributes.rb
```diff
@@ -26,12 +26,16 @@ def call(params)
...
-          Time.new(*args)
+      begin
+        if hour.blank? && minute.blank?
+          Date.new(year.to_i, month.to_i, day.to_i) # TODO: test fails.
+        else
+          args = [year, month, day, hour, minute].map(&:to_i)
+          Time.zone ? Time.zone.local(*args) :
+            Time.new(*args)
+        end
+      rescue ArgumentError
```
apotonick: Where exactly is the `ArgumentError` raised? Can we wrap only that part in the `rescue` block?

joealba: Both Date.new and Time.new could throw the ArgumentError on invalid input, so that's why I wrapped both in one shot.  Do you have another preference?

apotonick:
Fixed via #289. Thanks @joealba we need to have :beers: soon, again!

joealba:
:+1:

PR #268, opened 2015-07-21
Make `valid?` private

spectator:
Per conversion in #263 

`valid?` is widely used method in Rails and Lotus apps whereas in Reform
this method is undocumented, is not a part of public interface, and
should be used only in Reform internals.

Making `valid?` private should raise awareness to developers that they
might be calling something they shouldn't.

### DIFF
--- a/lib/reform/form/active_model/validations.rb
+++ b/lib/reform/form/active_model/validations.rb
@@ -66,6 +66,8 @@ module Reform::Form::ActiveModel
       end
     end
 
+    private
+
     # Needs to be implemented by every validation backend and implements the
     # actual validation. See Reform::Form::Lotus, too!
     def valid?
--- a/lib/reform/form/lotus.rb
+++ b/lib/reform/form/lotus.rb
@@ -42,14 +42,15 @@ module Reform::Form::Lotus
     end
   end
 
+  def build_errors
+    Errors.new
+  end
+
+  private
 
   def valid?
     # DISCUSS: by using @fields here, we avoid setters being called. win!
     validator = Lotus::Validations::Validator.new(self.class.validations, @fields, errors)
     validator.validate
-  end
-
-  def build_errors
-    Errors.new
   end
 end


fernandes:
how use should matchers with Reform?

How do I test if a contract is invalid under trailblazer? I'm used to operation.contract.valid?

Sorry to ask here on the PR, but I was tracking the issues and get here hahaha

thank you

spectator:
Here is an example from Trailblazer book which I highly recommend to read

```ruby
it "invalid email" do
  res, operation = Comment::Create.run(
    comment: { user: { email: "1337@" } }
  )

  res.must_equal false
  operation.errors.messages[:"user.email"].must_equal(["is invalid"])
end
```

or if you test your contract as stand-alone thing, you could just use its api in your tests the same way you use it in operations

rspec example

```ruby
describe MyContract do
  subject { described_class.new(MyModel.new) }
  before  { subject.validate({}) }

  it do
    expect(subject.errors[:my_attr]).to include("can't be blank")
  end
end
```

fernandes:
thank you @spectator , was thinking on how to test using the released API on v2

@apotonick nickitito, why break the API? Reform 2 was released weeks ago, could at least deprecated the #valid? :crying_cat_face:

apotonick:
@fernandes Celsito, `#valid?` is still there (and will stay there), you can use `#send`. It is incorrect to use contract internals in an operation test. You should use `operation.valid?` instead! :caipirinhas:

fernandes:
yeah... but shoulda-matchers relies on `#valid?` and I suspect that other rails stuff too (I confirm once I got it)

we do fuckin' need this emoji :caipirinhas:  hahahah

apotonick:
But `valid?` is simply the wrong entry point in Reform. 

The validation flow in Reform and "Rails" is different on purpose: A form is not a long-lived "model" that you change and validate and change and validate, it is a _"use once"_ throw-away object where there is no place for a public `valid?` method.

Changing the API just to make it testable with Rails-specific helpers is exactly what DHH talks about when he says TDD sucks. You never call `#validate` twice on a form in production code, so why would we allow that in tests?

I am still unsure if I like the matcher approach at all... in the book, we use full-blown `#validate` tests that are way closer to production-like data than isolated matcher tests, so.... I don't know... if you want to use matchers, you can make `valid?` public, but I really discourage this approach and matchers in general. :wefuckinneedcaipis:

fernandes:
yeah, `#validate` is called just one time, then I call `#valid?` like any controller action would do... I don't think the test diverges from production

I got your point, I'm gonna change the whole test suite to check if op is valid and is errors where added (for me a great huge step for a minor reform update 2.0.2 -> 2.0.3) :kissing_heart: 

:caipisftw!!:

apotonick:
Yes, I do not understand.

Where does "any controller action" call `valid?`. Do you have an example why you need this `valid?` call instead of the `validate` call?

This was a _bug_ that was fixed, the `valid?` method was planned to be private from 2.0 onwards. Sorry sweetheart! :kiss:

apotonick:
(Perfect excuse, from now on I will always say "this was a bug I had to fix it without a major version bump).

BTW, you can use `form.send(:valid?)`, the method will stay but it's not a public concept, if you need it that badly.

fernandes:
never mind, I'm gonna change here on my tests

perfect excuse, great hahahaha

apotonick:
So you gonna change the tests the way they are done in the book? Details!!!! :beers:

fernandes:
> Where does "any controller action" call valid?. Do you have an example why you need this valid? call instead of the validate call?

see page # 15 of your book :stuck_out_tongue_winking_eye: 

thats what I said, its "normal" behavior

PR #271, opened 2015-07-22
Use I18n for error message in uniqueness validator

yabawock:
Add :taken symbol as error to form if uniqueness validation fails. This
mimics the uniqueness validation from active model, making translation
of the error message possible, inheriting the default translations from
active_model / rails.

### DIFF
--- a/lib/reform/form/validation/unique_validator.rb
+++ b/lib/reform/form/validation/unique_validator.rb
@@ -10,7 +10,7 @@ class Reform::Form::UniqueValidator < ActiveModel::EachValidator
     query = query.merge(model.class.where("id <> ?", model.id)) if model.persisted?
 
     # if any models found, add error on attribute
-    form.errors.add(attribute, "#{attribute} must be unique.") if query.any?
+    form.errors.add(attribute, :taken) if query.any?
   end
 end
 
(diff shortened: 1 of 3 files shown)


yabawock:
Failed tests seem to be a travis issue - can't reproduce the SQlite errors on my end...

yabawock:
@apotonick Any update on this, would be really nice to have translated validation messages.

apotonick:
Does that still work against master?

yabawock:
Yeah still works, needed to explicitly add the translations to the test app to make the tests green on Rails < 4.0.

apotonick:
Cool, and `:taken` is the "official" way of handling uniqueness validations?

yabawock:
Yes, you can have a look at https://github.com/rails/rails/blob/master/activerecord/lib/active_record/validations/uniqueness.rb#L34 - it's been the same symbol since at least rails 3.0.

audionerd:
This worked well for me. 

One (possibly unrelated) thing that tripped me up: With AR's `validate uniqueness:true`, I had translations on the `activerecord` key in `en.yml`, but with I had to switch my translations for attribute name and error message to key off `activemodel` instead to get this patch to work.

apotonick:
@audionerd The `activemodel` "namespace" is because of the way I abstract `AM::Validations` from the form object. I might make it configurable.

apotonick:
Oh, and thanks you two (especially @yabawock) for your help! :heart:

Issue #249, opened 2015-07-15
broken of save with block


niedhui:
codes from README
```ruby
 @form.save do |hash|
    album = @form.model
    album.update_attributes(hash[:album])
  end
```
before 2_0, it works, the `hash` is `ActiveSupport::HashWithIndifferentAccess`. and now, the  `hash` is just a plain hash with string keys. we can only access it using `hash['album']`. Will this change back to the `AS::HWIA` style or  fix the README(and my codes~).
And I did not read the `disposable` codes, is it safe to just using the string keys, or I should cast the hash myself.

but when using composition, the hash will be `{:symbol_key => {"string_key" => "value"}}`， a litte 'not-unifiable-interface'

gertfj:
+1 for this, I have currently changed my code from:

```ruby
form.save do |hash|
  result = Model.create_model!(user: current_user, data: hash)
end
```

to

```ruby
form.save do |hash|
  result = Model.create_model!(user: current_user, data: hash.with_indifferent_access)
end
```

but it would be nice not having to remember doing that.

apotonick:
We could add the `.with_indifferent_access` but then only in `Form::ActiveRecord` forms. I stripped all the Rails-specific code from Reform in 2.0 and it feels great. So, we can re-introduce this but only for Rails forms.

gertfj:
Would be great, habits form quickly and I'm certain that I will forget to add it later.

apotonick:
Pull requests highly appreciated. I don't use `#save` with block so it might take a while. :wink:

PR #279, opened 2015-07-27
cast block hash to HWIA for active_record form

niedhui:
fix issue for #249
BTW,  currently(the master branch) the hash for block params is: 
1. symbol access key for model of  composition form
2. string access key for attribute
would you mind to change it all to string key or symbl key, when no acitve_record, no AS:HWIA

### DIFF
--- a/lib/reform/form/active_record.rb
+++ b/lib/reform/form/active_record.rb
@@ -20,6 +20,10 @@ module Reform::Form::ActiveRecord
     end
   end
 
+  def to_nested_hash(*)
+    super.with_indifferent_access
+  end
+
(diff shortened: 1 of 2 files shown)


apotonick:
You're right, all keys must be strings in the hash. Do you want to change that (is that only for `Composition`)?

niedhui:
I found it depends on how we create the `Twin` Composition  Object. Maybe we should just doc it 
```ruby
r1 = Request.new('song' => song, 'requester' => requester) 
r1.save do |hash|
  hash['song']
end
r2 = Request.new(song: song, requester: requester) 
r1.save do |hash|
  hash[:song]
end
```

PR #289, opened 2015-09-03
Fix invalid date issue without catch or raise exception

halan:
This solve https://github.com/apotonick/reform/pull/267 without catch generic exceptions or any other. I use the same test at the other PR.

### DIFF
--- a/lib/reform/form/multi_parameter_attributes.rb
+++ b/lib/reform/form/multi_parameter_attributes.rb
@@ -24,14 +24,17 @@ module Reform::Form::MultiParameterAttributes
 
   private
     def params_to_date(year, month, day, hour, minute)
-      if [year, month, day].any?(&:blank?) || !Date.valid_date?(*[year, month, day].map(&:to_i))
+      date_fields = [year, month, day].map!(&:to_i)
+      time_fields = [hour, minute].map!(&:to_i)
+
+      if date_fields.any?(&:zero?) || !Date.valid_date?(*date_fields)
         return nil
       end
 
       if hour.blank? && minute.blank?
-        Date.new(year.to_i, month.to_i, day.to_i)
+        Date.new(*date_fields)
       else
-        args = [year, month, day, hour, minute].map(&:to_i)
+        args = date_fields + time_fields
         Time.zone ? Time.zone.local(*args) :
           Time.new(*args)
       end
--- a/lib/reform/form/multi_parameter_attributes.rb
+++ b/lib/reform/form/multi_parameter_attributes.rb
@@ -24,10 +24,12 @@ module Reform::Form::MultiParameterAttributes
 
   private
     def params_to_date(year, month, day, hour, minute)
-      return nil if [year, month, day].any?(&:blank?)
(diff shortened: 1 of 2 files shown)


PedroFelipe:
:+1:

matheusazzi:
:+1:

akaKuruma:
:+1:

luizrogeriocn:
:+1:

cefigueiredo:
:+1:

brunoskonrad:
:+1:

iagomoreira:
smooth :+1:

bernardogfilho:
nice :+1:

luizvarela:
:+1:

romulostorel:
:+1:

halan:
:+1:

lib/reform/form/multi_parameter_attributes.rb
```diff
@@ -24,10 +24,12 @@ def call(params)
 
   private
     def params_to_date(year, month, day, hour, minute)
-      return nil if [year, month, day].any?(&:blank?)
+      if [year, month, day].any?(&:blank?) || !Date.valid_date?(*[year, month, day].map(&:to_i))
```
carloslopes: You are generating the same array twice, can you refactor and reuse the same?

waldyr:
:-1:

lib/reform/form/multi_parameter_attributes.rb
halan: I think about it, I just try solve the problem with minimum code, but I will do it, above have the same array too.

lib/reform/form/multi_parameter_attributes.rb
```diff
@@ -24,10 +24,12 @@ def call(params)
...
+        return nil
+      end
 
       if hour.blank? && minute.blank?
```
carloslopes: if you are handling dates, why are you concerned about hour and minute?

lib/reform/form/multi_parameter_attributes.rb
andrewhr: Blank strings become `0` when coerced via `to_i`. Given that any `0` is understood as an invalid date, what do you think of skipping the `blank?` check? Maybe I'm missing something?

apotonick:
Haha, great but since when do we have this :+1: on my gems?! :laughing:

apotonick:
Cool, so this catches invalid dates without an exception? That's great! Any concerns?

halan:
No, it's ruby 1.8 compatible. It will return `nil` if receives invalid date, instead `ArgumentError` raised by `Date.new` or `Time.new`

halan:
I just put the tests from #267.

apotonick:
@halan But here you test invalid dates: https://github.com/apotonick/reform/pull/289/files#diff-ad5370b7e7280e5aeb58a9412a4ade3bR20 and I can't see any `rescue` in your code.

Please join https://gitter.im/trailblazer/chat for a quick chat! :beers:

brunoskonrad:
:clap:

Issue #296, opened 2015-10-15
Multi-parameter dates fail in Rails 5


audionerd:
Not sure what's causing this, although I'm guessing it's because I'm on Rails 5.

When following these instructions:

https://github.com/apotonick/reform#multiparameter-dates

I get this error:

```ruby
*** NoMethodError Exception: undefined method `merge!' for #<ActionController::Parameters:0x008fec01e74f60>
```

It's triggered by `reform/lib/reform/form/multi_parameter_attributes.rb`

https://github.com/apotonick/reform/blob/c2db7e661d0d5e8de208c8aadc211e8c8b3a8527/lib/reform/form/multi_parameter_attributes.rb#L22

If I swap

```ruby
params.merge!(date_attributes)
```

with:

```ruby
params = params.merge(date_attributes)
```

... then it works!

ActionController::Parameters#merge appears to be the way to go for Rails 5:
https://github.com/rails/rails/blob/0450642c27af3af35b449208b21695fd55c30f90/actionpack/lib/action_controller/metal/strong_parameters.rb#L563-L569

audionerd:
Updated Rails to the latest master and I no longer need this fix.

PR #297, opened 2015-10-18
Fix multi param attributes for Rails 5

audionerd:
Fixes https://github.com/apotonick/reform/issues/296. Probably worth some discussion – this fixed "worked for me" but there might be a better way to solve the issue.


audionerd:
Updated Rails to the latest master and I no longer need this fix.

PR #300, opened 2015-10-23
Add scope options to unique validation

acaron:
Simple scope option! Yey!

To use it:
`validates :title, unique: { scope: :album_id }`

It checks only song with the same title by album instead of all the songs available.


apotonick:
This is great! :heart:

acaron:
Quick note: you can do `validates :title, unique: { scope: [:album_id, :artist_id] }` also!

PR #304, opened 2015-10-26
Validate unique inside an association that is persisted or not

acaron:
The unique validation must check if the property to validate is a collection or a field. If it is a collection, it use the scope to compare each fields. Very useful when you have nested fields inside the same form where the user can add has many nested items that he wants. The nested fields are not yet save in the database so it must be check in another way then a query.

Thank you to: http://stackoverflow.com/questions/2772236/validates-uniqueness-of-in-destroyed-nested-model-rails

```
class AlbumForm < Reform::Form
    property :title
    validates :songs, unique: { scope: :title }

    collection :songs, :populate_if_empty => Song do
      property :title
    end
    [...]
```

It doesn't check `marked_for_destruction?` since `_destroy` is not yet support in `Reform`. Do you think it must be add if someone manually decide to call: `marked_for_destruction`? (`if key.blank? || record.try(:marked_for_destruction?)` instead on line 15)

### DIFF
--- a/lib/reform/form/validation/unique_validator.rb
+++ b/lib/reform/form/validation/unique_validator.rb
@@ -1,22 +1,45 @@
 # Reform's own implementation for uniqueness which does not write to model.
 class Reform::Form::UniqueValidator < ActiveModel::EachValidator
   def validate_each(form, attribute, value)
+    has_error = false
     model = form.model_for_property(attribute)
+    scope = Array(options[:scope])
 
-    # search for models with attribute equals to form field value
-    query = model.class.where(attribute => value)
+    if value.is_a?(Array)
+      # validate in memory collection
+      # thank you: http://stackoverflow.com/questions/2772236/validates-uniqueness-of-in-destroyed-nested-model-rails
+      hashes = value.inject({}) do |hash, record|
+        key = scope.map { |a| record.send(a).to_s }.join
 
-    # apply scope if options has been declared
-    Array(options[:scope]).each do |field|
-      # add condition to only check unique value with the same scope
-      query = query.where(field => form.send(field))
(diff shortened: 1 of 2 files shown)


acaron:
Same here, build don't pass but PR does

meaganewaller:
Is this still something that we'd like to see merged into Reform? If so, is there anything I can do to help this get across the finish line?

apotonick:
We should extract this unique validator to a separate object that is decoupled from Rails/ActiveRecord's API. This validator is starting to get really complex, but I can see how that helps.

The goal is to make this validator available for other validation engines, e.g. our `dry-validation` integration, since we drop (active) support for `ActiveModel::Validations` in Reform 2.2.

I am picturing something as follows.

```ruby
class AnnieAndMeagansUniqueValidator # NO inheritance here
  def validate(form, attribute, value)
    # here goes all the logic for scopes, etc. this can be nicely architected within this class
  end
end
```

We can then integrate that with all validation backends, *and*, if needed, with different ORMs. The other cool thing is: we can run all kinds of unit tests against this isolated class and don't need a form setup while working on it.

Actually, what about a gem `reform-validators` that holds those? Would you ladies like to work on that? I can help with the initial setup/test structure.

meaganewaller:
That sounds like a great solution! I'd love to help work on that once the initial setup is in place.

acaron:
@apotonick this `reform-validators` is interesting! I decided to use Rails `uniqueness` validator for now because I don't want race conditions to happen. Are you willing to add adapters to it so it supports Rails or Lotus, etc. database errors? Rails throws `ActiveRecord::StatementInvalid` for example (https://github.com/rails/rails/blob/master/activerecord/lib/active_record/validations/uniqueness.rb).

fran-worley:

PR #308, opened 2015-11-04
Allows Sequel models to use UniqueValidator by removing call to query.merge,...

amulter:
...which Sequel choked on.

Also removes dependency on model.id for tables with differently named primary keys.


apotonick:
Looks like a very clever fix even though I have no idea if it'll work! :stuck_out_tongue_winking_eye: Thanks!

PR #309, opened 2015-11-04
cache model_name

timoschilling:


apotonick:
I hate this module so much that I don't know if adding variable caching will make it any better! :stuck_out_tongue_winking_eye:

Issue #230, opened 2015-06-08
Problem with copy_validations_from


Szeliga:
Hi,

As stated in the title, I can't get it to work. I have a User class

```ruby
class User < ActiveRecord::Base
  # Include default devise modules. Others available are:
  # :confirmable, :lockable, :timeoutable and :omniauthable
  devise :database_authenticatable, :registerable,
         :recoverable, :rememberable, :trackable, :validatable

  validates :name, length: {minimum: 10}
end
```

And a form:

```ruby
module Forms
  class Registration < Reform::Form
    extend ActiveModel::ModelValidations

    copy_validations_from User

    properties :email, :password, :name, :location, validates: { presence: true }
    property :password_confirmation, empty: true

    validates :email, format: Devise.email_regexp, uniqueness: true
    validates :password, confirmation: true
    validates :password, length: { within: Devise.password_length }

    model :user
  end
end
```

When I try to validate the form, it skips the length validation on the name property:

```
2.2.2 :001 > form = Forms::Registration.new(User.new)
 => #<Forms::Registration:0x007fa0823d1718 @model=#<User id: nil, email: "", encrypted_password: "", reset_password_token: nil, reset_password_sent_at: nil, remember_created_at: nil, sign_in_count: 0, current_sign_in_at: nil, last_sign_in_at: nil, current_sign_in_ip: nil, last_sign_in_ip: nil, created_at: nil, updated_at: nil, name: nil, location: nil>, @fields=#<Reform::Contract::Fields email="", password=nil, name=nil, location=nil, password_confirmation=nil>>
2.2.2 :002 > form.validate(email: 'foo@bar.com', password: 'foobar12', password_confirmation: 'foobar12', name: 'foobar', location: 'Neverland')
  User Exists (6.4ms)  SELECT  1 AS one FROM "users" WHERE "users"."email" = 'foo@bar.com' LIMIT 1
 => false
2.2.2 :004 > form.errors.messages
 => {:email=>["has already been taken"]}
```

Am I doing something wrong here?

Best regards, Szymon Szeliga

apotonick:
I don't think so, but I didn't write the `copy_validations` feature so I'm not sure I'll be of great help here... Is it only one validation that's not being copied, or all?

Szeliga:
I've only tested it on a model with one validation, but I will check for more

Szeliga:
[...]

PR #321, opened 2015-12-02
Updated Readme for copy_validations_from

francirp:
This Readme update is driven by the following issue: https://github.com/apotonick/reform/issues/230

It is unreasonable to expect people to keep the copy_validations_from line under the property definitions. We should explicitly let people know this structure is required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -734,7 +734,7 @@ class SongForm < Reform::Form
 end
 ```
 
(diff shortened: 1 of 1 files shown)


apotonick:
Oh good one, can you squash those 3 into one commit? Thanks! :heart:

francirp:
@apotonick ah good call  - squashed em'

apotonick:
Rock'n'roll! Thanks!

PR #324, opened 2015-12-10
Featuring dry-validations initial support!

mrbongiolo:
Basic stuff seems to be working.

There's still a bug with nested validations, the nested `validation` blocks thinks he's in an AM:V party, not the Dry::Schema we used in the main form =/ @apotonick said he probably knows why that's happening.

The `validation` block is executed in the context of Dry::Schema, so everything that's possible in it can be done in the block :D (setting `config` and all that stuff works o/).
Note that this way we can't use `validate`, `validates`, `validates_with` nor any kind of `validate` directly from the property. I think it's fine that way, since that's how dry-v is supposed to work.

Whoever wants to use this will understand that they should use `key` and all the dry-v predicates xD


robjacoby:
:+1: @apotonick gogogo! :wink:

Issue #313, opened 2015-11-09
collections with :skip_if filled with nil, raises exception on validate


akamike:
A collection using `skip_if` becomes filled with `nil` when skipped items are before non-skipped items, due to retaining the original indexes. While that particular setup seems to be the responsibility of [Representable](https://github.com/apotonick/representable), it can cause issues in reform - it shouldn't be trying to call `validate!` on `nil`.

This is not caught by the [tests](https://github.com/apotonick/reform/blob/master/test/skip_if_test.rb#L43) because it only tests for skipped items at the end of the input. My expectation would be for the following addition to pass (it currently does not):

````ruby
# additions to test/skip_if_test.rb
class SkipIfTest < BaseTest
  # ...

  # skips deserialization when not present, regardless of position
  it do
    form = AlbumForm.new(Album.new(nil, nil, []))
    form.validate("songs" => [
      {"title" => nil},
      {"title" => "Altar Of Sacrifice"},
      {"title" => nil},
      {"title" => "Waste Of Breath"},
      {"title" => nil},
      {"title" => "Fear of a Blank Planet"},
      {"title" => nil}
    ]).must_equal true
    form.songs.size.must_equal 3
    form.songs[0].title.must_equal "Altar Of Sacrifice"
    form.songs[1].title.must_equal "Waste Of Breath"
    form.songs[2].title.must_equal "Fear of a Blank Planet"
  end
end
````

The fix for this is to reset the indexes on the collection, since just handling this in validate causes issues with the returned collection. However, I noticed a few other tests that seem to rely on this behaviour:

* https://github.com/apotonick/reform/blob/master/test/deprecation_test.rb#L24
* https://github.com/apotonick/reform/blob/master/test/---populator_test.rb#L23
* https://github.com/apotonick/reform/blob/master/test/populate_test.rb#L199 (this is [present on 2.0.5](https://github.com/apotonick/reform/blob/v2.0.5/test/populate_test.rb#L190))

Raising this as an issue to help clarify what is the expected behaviour, and perhaps look at ways to accomplish my provided test case. `deprecation_test.rb` and `---populator_test.rb` seem to offer a workaround for the validation part, but this feels hacky and incomplete.

kevinbrink:
I believe this is the exact same issue I've been struggling with for the past few days. 

As a side note, I'm also using [cocoon](https://github.com/nathanvda/cocoon), but I don't think that's causing any issues. 

[...]

PR #328, opened 2015-12-17
Tries to [Fix] #313 on upstream branch

kevinbrink:
- Resolve issue with adding and removing items in a collection in a single action

The app I'm working on is still using version 2.0.5 of reform, so my fork is only up to date since [this commit](https://github.com/apotonick/reform/commit/16ab6083926488b65b6232262ae69e2d7945528b), hence the merge conflicts. 

It would be nice to get this change incorporated into version 2.1, when that gets released. The most up to date version of the reform branch (something like 2.1.pre1, I think) doesn't work properly with the app, and I don't have time right now to figure out how to make it compatible. 

There are two skipped tests, but other than that, they all pass based on version 2.0.5.

### DIFF
--- a/lib/reform/form/populator.rb
+++ b/lib/reform/form/populator.rb
@@ -50,7 +50,7 @@ private
         index = args.first
         item = twin.original[index] and return item
 
-        twin.insert(index, run!(form, fragment, options)) # form.songs.insert(Song.new)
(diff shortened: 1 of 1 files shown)


apotonick:
https://github.com/apotonick/reform/commit/c7a1bc247cf66c95ef2c269f77f2bb5f32470170 :heart:

PR #333, opened 2016-01-19
Expand uniqueness handling.

spectator:
Handle uniqueness case when updating existing record attribute to the
value of another existing record attribute.

---

I run into issue when updating an attribute of existing record to the value of another existing record doesn't trigger uniqueness validation error. I adjusted the condition based on what it is being decided is it is not unique so that it checks we are updating the same record and in that case we are OK that count will return 1. In all other cases count should return 0 to make sure value is unique.

### DIFF
--- a/lib/reform/form/validation/unique_validator.rb
+++ b/lib/reform/form/validation/unique_validator.rb
@@ -41,8 +41,9 @@ class Reform::Form::UniqueValidator < ActiveModel::EachValidator
       query = query.where(field => form.send(field))
     end
 
-    # if model persisted, query may return 0 or 1 rows, else 0
-    allow_count = model.persisted? ? 1 : 0
+    # if model is persisted and has the same value then count of 1 is allowed, in
+    # all other other cases it is 0
+    allow_count = model.persisted? && model.public_send(attribute) == value ? 1 : 0
     form.errors.add(attribute, :taken) if query.count > allow_count
   end
 end
--- a/test/unique_test.rb
+++ b/test/unique_test.rb
@@ -42,8 +42,16 @@ class UniquenessValidatorOnUpdateTest < MiniTest::Spec
     form = SongForm.new(@song)
     form.validate("title" => "How Many Tears").must_equal true
   end
-end
 
+  it do
+    Song.delete_all
+    Song.create(title: "How Many Tears")
+
+    @song = Song.create(title: "Baby Face")
+    form = SongForm.new(@song)
+    form.validate("title" => "How Many Tears").must_equal false
+  end
+end
 
 class UniqueWithCompositionTest < MiniTest::Spec
   class SongForm < Reform::Form


spectator:
I just realized that implementation is a little too Rails-ish. I guess `model[attribute]` should be something like `model.public_send(attribute)` ?

apotonick:
I don't really understand why that works, but OK (the `1` count)! And, yes, a `public_send` will be less AR and will probably work with all ORMs.

spectator:
Updated.

Here is why it doesn't work in all cases. https://github.com/apotonick/reform/blob/master/lib/reform/form/validation/unique_validator.rb#L45 says that count query might return 0 or 1 for persisted model.

Obviously, it returns 1 if record already exist, however that `1` has different meaning in different conditions. 

Consider the following table

|id|name
---|--------
|1 |a
|2 |b

If form is working with record 2 and tries to set the name to 'a', count query would return 1, because in this case it means record will hot be unique.

If form is working with record 1 and tries to set the name to 'a', count query would return 1, but this is fine because we are updating the same record to the same value, so count of 1 in this case means it still will be unique. **This is what I enforce in this PR**, before that this condition wasn't string enough.

Maybe there is less confusing way. Let's say check that current model is the same record when there is just one... maybe some other way?

spectator:
Maybe just change this query

```ruby
query = model.class.where(attribute => value)
```

to 

```ruby
query = model.class.where(attribute => value).where.not(id: model.id)
```

?

Although I'm not sure if that is better and how to write it so it is compatible with different ORMs.

spectator:
On a side note, this validation is actually AR specific, ins't it? So it has to work with just one ORM and probably should be in `reform-rails` gem?

spectator:
How do you like this? I think it reads way better

```diff
--- c/lib/reform/form/validation/unique_validator.rb
+++ w/lib/reform/form/validation/unique_validator.rb
@@ -33,7 +33,7 @@ class Reform::Form::UniqueValidator < ActiveModel::EachValidator
     model = form.model_for_property(attribute)
 
     # search for models with attribute equals to form field value
-    query = model.class.where(attribute => value)
+    query = model.class.where(attribute => value).where.not(id: model.id)
 
     # apply scope if options has been declared
     Array(options[:scope]).each do |field|
@@ -41,9 +41,7 @@ class Reform::Form::UniqueValidator < ActiveModel::EachValidator
       query = query.where(field => form.send(field))
     end
 
-    # if model persisted, query may return 0 or 1 rows, else 0
-    allow_count = model.persisted? ? 1 : 0
-    form.errors.add(attribute, :taken) if query.count > allow_count
+    form.errors.add(attribute, :taken) if query.count > 0
   end
 end
```

spectator:
Looked into rails and they do exactly the same thing https://github.com/rails/rails/blob/master/activerecord/lib/active_record/validations/uniqueness.rb#L21-L25 with some extra logic to fail if there is no primary key.

apotonick:
Yeah, but `where.not` won't work with most ORMs, whereas only `where` goes fine with most!

spectator:
That condition could be `.where("id <> ?", model.id)` and that might work with most ORMs too, right?

apotonick:
I remember we had that a while ago...

spectator:
So what would you suggest? Validator is clearly not doing what is supposed to do, how would you like this to be implemented? I'm running our of ideas :D

spectator:
So this PR https://github.com/apotonick/reform/pull/308 obviously introduced regression. It supposed to fix Sequel. Sequel now won't blow up, but validator won't work correctly for Sequel and any other ORM, it does't guarantee uniqueness now (on its level).

My suggestion would be to restore original behavior and then think of a more flexible way to support different ORMs than just sticking to `.where(hash)`. Maybe since this could be very specific to ORM then let everybody implement their own uniqueness validation and extract this one into `reform-rails` as an example?

apotonick:
Or `reform-orm` where we have different modules?

spectator:
Yep, or `reform-orm`

apotonick:
Actually, I had the idea of `reform-validators`. We can then have different implementations there.

spectator:
Yeah, I remember you mentioned this a while ago, surprised this is still not implemented :trollface:

apotonick:
Hahahaha ass!

fran-worley:

PR #340, opened 2016-03-02
Add :parse option for property

pabloh:
Allow defining *parse* properties that cannot be updated at `Form#validate`:

```ruby
User = Struct.new(:name, :admin)

class UserForm < Reform::Form
  property :name
  property :admin, parse: false
end

user = User.new("Peter", false)
form = UserForm.new(user)

form.validate("name" => "Joseph", "admin" => true)

form.name # returns "Joseph"
form.admin # returns false
```

### DIFF
--- a/lib/reform/form.rb
+++ b/lib/reform/form.rb
@@ -16,6 +16,11 @@ module Reform
     module Property
       # Add macro logic, e.g. for :populator.
       def property(name, options={}, &block)
+        if options.key?(:parse)
+          options[:deserializer] ||= {}
+          options[:deserializer][:writeable] = options.delete(:parse)
+        end
+
         definition = super # let representable sort out inheriting of properties, and so on.
         definition.merge!(deserializer: {}) unless definition[:deserializer] # always keep :deserializer per property.
 
--- a/test/parse_option_test.rb
+++ b/test/parse_option_test.rb
@@ -1,0 +1,40 @@
+require 'test_helper'
+
+class ParseOptionTest < MiniTest::Spec
+  Comment = Struct.new(:content, :user)
+  User    = Struct.new(:name)
+
+  class CommentForm < Reform::Form
+    property :content
+    property :user, parse: false
+  end
+
+  let (:current_user) { User.new("Peter") }
+  let (:form) { CommentForm.new(Comment.new, user: current_user) }
+
+  it do
+    form.user.must_equal current_user
+
+    lorem = "Lorem ipsum dolor sit amet..."
+    form.validate("content" => lorem, "user" => "not the current user")
+
+    form.content.must_equal lorem
+    form.user.must_equal current_user
+  end
+
+  describe "using ':parse' option doesn't override other ':deserialize' options" do
+    class ArticleCommentForm < Reform::Form
+      property :content
+      property :article, deserializer: { instance: "Instance" }
+      property :user, parse: false, deserializer: { instance: "Instance" }
+    end
+
+    it do
+      ArticleCommentForm.definitions.get(:user)[:deserializer][:writeable].must_equal false
+      ArticleCommentForm.definitions.get(:user)[:deserializer][:instance].must_equal "Instance"
+
+      ArticleCommentForm.definitions.get(:article)[:deserializer][:writeable].must_equal true
+      ArticleCommentForm.definitions.get(:article)[:deserializer][:instance].must_equal "Instance"
+    end
+  end
+end


fran-worley:
This has been incorporated into Reform 2.1.0 see the [docs](http://trailblazer.to/gems/reform/)

**Read-Only Fields**

When you want to show a value but skip processing it after submission the `:writeable` option is your friend.

```ruby
class ProfileForm < Reform::Form
  property :country, writeable: false
```

This time reform will query the model for the value by calling `model.country`.

You want to use this to display an initial value or to further process this field with JavaScript. However, after submission, the field is no longer considered: it won’t be written to the model when saving.

It is still readable in the nested hash and through the form itself.

```ruby
form.save do |nested|
  nested[:country] #=> "Australia"
```

pabloh:
@fran-worley , the problem about `writeable: false` is that it only prevents your form from writing back to the model when doing `sync`/`save`, what I wanted is to prevent the input value reaching the form altogether after calling `form.validate(params)`. 

I ask @apotonick and he pointed at me some section of the TBZ book, where I basically have to declare the property like this: `property :country, deserializer: {writeable: false}`, which was awkward and verbose. I wanted a shorter and more see-through option to have that same effect. (Also he proposed to use `readonly:`).

fran-worley:
@pabloh ahh that makes sense, thanks for the update. I'll reopen!

pabloh:
@fran-worley thank you :smile:.

apotonick:
This already works and is shown in the Trailblazer book. `deserializer: { writeable: false }`

apotonick:
Not sure, will the `:readonly` option be confusing?

fran-worley:
I was worried that people might get mixed up. im wondering whether the existing option is sufficient given how rarely it's used and that of its not a rare occurance could that be a sign that you need to revisit what your doing?

pabloh:
I don't have to use it that often, basically, it allows me to completely forgot about protecting some sensible attributes like you would by using `StrongParameters`. I feel there is a need for this feature to exists although I may concede that `readonly:` could be confusing since there is already `readable:`.

apotonick:
Now I'm confused. How is that related to `strong_parameters`? The reason we have `property` is to expose fields in a form ("graphical"). If you don't want them writeable, you define the field as `writeable: false`. If you don't want any of that field to be exposed, you don't define it as a property, and Reform will ignore it when deserializing the input. Is that what you mean?

fran-worley:
I'm going to side with @apotonick. If you don't want an value to end up in your form instance then you shouldn't define it as a property. If you need a reader on your form object, just create a method e.g.

```ruby
def some_attribute_i_need_access_to_but_dont_want_to_do_anything_with
  model.my_attribute_getter # or something else
end
```

That way your form can access it as if it was a property but you wont ever have it coming back as a param.

pabloh:
@apotonick , @fran-worley , maybe my analogy with `StrongParameters` was a bad one.

The most common use case I have for this is when I need to pass some contextual data to a contract initializer. For instance, some domain level model for an association (most of the time the current user). That property needs to be part of the form but I really don't want risk someone tampering the value from the HTTP request.

apotonick:
Ah, good point, @pabloh. Maybe an option `exclude_from_parsing: true` or parse: false` would clarify that?

fran-worley:
I have encountered this a few times but I tend to make them a reader on the form class rather than a property.

Don't get me wrong I'm not adverse to this, I just want to make sure that we are finding the correct way to deal with a use case rather than adding an additional option to fix an example .

I also think that read_only with confuse and that anything long enough not too and you might find that the existing functioning solution of a deserialise option is equally valid...


On 6 Jul 2016, at 23:06, Pablo Herrero <notifications@github.com<mailto:notifications@github.com>> wrote:


@apotonick<https://github.com/apotonick> , @fran-worley<https://github.com/fran-worley> , maybe my analogy with StrongParameters was a bad one.

The most common use case I have for this is when I need to pass some contextual data to a contract initializer. For instance, some domain level model for an association (most of the time the current user). That property needs to be part of the form but I really don't want risk someone tampering the value from the HTTP request.

-

pabloh:
@fran-worley , thank you for your answer. 
Regarding using a reader attr, it wouldn’t be what I need most of the time, since I usually want to actually assign the property to the model during the sync and a normal attribute reader would be ignored.
Also, I really don't back 100% the name `readonly`, maybe `parse: false`, or `protect: true` are better names?.

fran-worley:
Thinking about this I had a couple of examples where this might have been very useful. 

Happy to go with this if you can rename the option. I think I prefer parse false as (if I understand this correctly) a better description of what is happening.

pabloh:
I'm ok with using `parse: false`, @apotonick are you on board with that?.

pabloh:
@fran-worley, @apotonick, I've updated the PR to use `parse: false` instead.

apotonick:
I like that.

Can you add a test that makes sure the `:parse` option doesn't override other `:deserializer` options? You can simply test `Your::Form.definitions.get(:tested_property)[:deserializer]`! :beers:

pabloh:
@apotonick , ok, done!

apotonick:
Adding docs here: http://trb.to/gems/reform/api.html#access-protection

PR #346, opened 2016-03-10
Don't rely on AR column information being present for all properties

niels:
I was having trouble creating a nested form based off of a property that contains Hash-like data. simple_form tried calling the AR-specific #column_for_attribute method on that property. Reform blindly delegated that method to the underlying object – which however did not provide it, as it wasn't an AR object.

The attached patch only delegates if the method is actually defined. Otherwise, nil is returned. I believe that this is actually the intent behind simple_form's code as well, [as it checks #respond_to? before calling](https://github.com/plataformatec/simple_form/blob/e81af052f4bbe9e17e38c668116cd1846c6d275c/lib/simple_form/form_builder.rb#L547). Reform broke that functionality by *always* responding.


niels:
(The only change to database.sqlite3 is the addition of a "meta" column to the Song schema.)

apotonick:
The idea here was to only include the `ModelReflections` module (which allows SimpleForm to query meta data from columns) if the form is backed by an AR model. If not, simply don't include the module. That way, we can save `respond_to?` as it makes things unnecessarily complex and has made my life hell many times (because Rails does this *everywhere*).

I am guessing you're using `feature ModelReflections` which will automatically include the module in all nested forms. Use `include ModelReflections` instead for a per-form inclusion.

niels:
Mmmh, I actually `include` it into `Reform::Form` itself.

The reason is simple: because (a) 99% of my Reform at this point objects *are* backed by an AR model, making the ModelReflections inclusion a reasonable default and (b) only a lazy developer is a good developer.

From that standpoint, it would be nice if things like AR reflections could be included blindly, knowing that they turn themselves off or on automatically as appropriate. Nevertheless, I take your point. I guess an alternative solution would then be for me to create a `Reform::Form::ActiveRecord` subclass for my AR-backed forms and leave `Reform::Form` pristine.

Regardless of all of the above however, none of this works in practice because irrespective of nesting, you can't `include` without registering the `feature` due to https://github.com/apotonick/reform/blob/master/lib/reform/form/active_model/model_reflections.rb#L10 :stuck_out_tongue_closed_eyes:

I.e. even the below leads to the reflections being included in the nested form:

```ruby
require "reform/form/active_model/model_reflections"

class MyForm < Reform::Form
  include Reform::Form::ActiveModel::ModelReflections

  property :a_struct do
    include Disposable::Twin::Struct
  end
end

my_form = MyForm.new(OpenStruct.new(a_struct: OpenStruct.new))
puts my_form.respond_to?(:column_for_attribute).inspect           #=> "true" – Good!
puts my_form.a_struct.respond_to?(:column_for_attribute).inspect  #=> "true" – Bad!
```

apotonick:
Argh, I hate `ModelReflections` and I hope with the upcoming Formular gem, we won't need SimpleForm and its annoying magic anymore. 

A simple work-around for now would be to `undef` the `#column_for_attribute` method in the particular form. I am not gonna put more time into this as I am not interested in Active* anymore. 

This "bug" is a great example why `respond_to?` is crap and should *never* be used as a tool in a framework. I had a brief discussion with @tenderlove about this but unfortunately it died down before we even got started.

PR #347, opened 2016-03-10
Integrate new dry-validation DSL

blelump:
Hi @apotonick ,

this PR is stil in very premature version since `dry-v` 0.7.0 haven't been released yet. However, it integrates the new DSL, which will be released with the upcoming version so let's test it! :-)

### DIFF
--- a/test/validation/dry_validation_test.rb
+++ b/test/validation/dry_validation_test.rb
@@ -46,20 +46,20 @@ class ValidationGroupsTest < MiniTest::Spec
     # invalid.
     it do
       form.validate({}).must_equal false
-      form.errors.messages.inspect.must_equal "{:username=>[\"username must be filled\"], :email=>[\"email must be filled\"]}"
+      form.errors.messages.inspect.must_equal "{:username=>[\"username is missing\"], :email=>[\"email is missing\"]}"
     end
 
     # partially invalid.
     # 2nd group fails.
     it do
       form.validate(username: "Helloween", email: "yo", confirm_password:"9").must_equal false
-      form.errors.messages.inspect.must_equal "{:email=>[\"email size cannot be less than 3\"], :confirm_password=>[\"confirm_password size cannot be less than 2\"], :password=>[\"password must be filled\", \"password size cannot be less than 2\"]}"
+      form.errors.messages.inspect.must_equal "{:email=>[\"email size cannot be less than 3\"], :confirm_password=>[\"confirm_password size cannot be less than 2\"], :password=>[\"password is missing\", \"password size cannot be less than 2\"]}"
     end
     # 3rd group fails.
     it do
       form.validate(username: "Helloween", email: "yo!", confirm_password:"9").must_equal false
       form.errors.messages.inspect
-      .must_equal "{:confirm_password=>[\"confirm_password size cannot be less than 2\"], :password=>[\"password must be filled\", \"password size cannot be less than 2\"]}"
+      .must_equal "{:confirm_password=>[\"confirm_password size cannot be less than 2\"], :password=>[\"password is missing\", \"password size cannot be less than 2\"]}"
     end
     # 4th group with after: fails.
     it do
@@ -96,6 +96,16 @@ class ValidationGroupsTest < MiniTest::Spec
       end
 
       validation :default do
+
+        key(:title).required
+
+        key(:band).schema do
+          key(:name).required
+          key(:label).schema do
+            key(:name).required
+          end
+        end
+
         configure do
           option :form
           # message need to be defined on fixtures/dry_error_messages
@@ -245,7 +255,7 @@ class ValidationGroupsTest < MiniTest::Spec
     # invalid.
     it do
       form.validate({}).must_equal false
-      form.errors.messages.inspect.must_equal "{:email=>[\"email must be filled\"], :username=>[\"username must be filled\"]}"
+      form.errors.messages.inspect.must_equal "{:email=>[\"email is missing\"], :username=>[\"username is missing\"]}"
     end
   end
 
@@ -283,7 +293,7 @@ class ValidationGroupsTest < MiniTest::Spec
     # invalid.
     it do
       form.validate({email: 9}).must_equal false
-      form.errors.messages.inspect.must_equal "{:username=>[\"username must be filled\"]}"
+      form.errors.messages.inspect.must_equal "{:username=>[\"username is missing\"]}"
     end
   end
 
--- a/test/validation/dry_validation_test.rb
+++ b/test/validation/dry_validation_test.rb
@@ -46,25 +46,25 @@ class ValidationGroupsTest < MiniTest::Spec
     # invalid.
     it do
       form.validate({}).must_equal false
-      form.errors.messages.inspect.must_equal "{:username=>[\"username is missing\"], :email=>[\"email is missing\"]}"
+      form.errors.messages.inspect.must_equal "{:username=>[\"is missing\"], :email=>[\"is missing\"]}"
     end
 
     # partially invalid.
     # 2nd group fails.
     it do
       form.validate(username: "Helloween", email: "yo", confirm_password:"9").must_equal false
-      form.errors.messages.inspect.must_equal "{:email=>[\"email size cannot be less than 3\"], :confirm_password=>[\"confirm_password size cannot be less than 2\"], :password=>[\"password is missing\", \"password size cannot be less than 2\"]}"
+      form.errors.messages.inspect.must_equal "{:email=>[\"size cannot be less than 3\"], :confirm_password=>[\"size cannot be less than 2\"], :password=>[\"is missing\", \"size cannot be less than 2\"]}"
     end
     # 3rd group fails.
     it do
       form.validate(username: "Helloween", email: "yo!", confirm_password:"9").must_equal false
       form.errors.messages.inspect
-      .must_equal "{:confirm_password=>[\"confirm_password size cannot be less than 2\"], :password=>[\"password is missing\", \"password size cannot be less than 2\"]}"
+      .must_equal "{:confirm_password=>[\"size cannot be less than 2\"], :password=>[\"is missing\", \"size cannot be less than 2\"]}"
     end
     # 4th group with after: fails.
     it do
       form.validate(username: "Helloween", email: "yo!", password: "", confirm_password: "9").must_equal false
-      form.errors.messages.inspect.must_equal "{:confirm_password=>[\"confirm_password size cannot be less than 2\"], :password=>[\"password must be filled\", \"password size cannot be less than 2\"]}"
+      form.errors.messages.inspect.must_equal "{:confirm_password=>[\"size cannot be less than 2\"], :password=>[\"must be filled\", \"size cannot be less than 2\"]}"
     end
   end
 
@@ -255,7 +255,7 @@ class ValidationGroupsTest < MiniTest::Spec
     # invalid.
     it do
       form.validate({}).must_equal false
-      form.errors.messages.inspect.must_equal "{:email=>[\"email is missing\"], :username=>[\"username is missing\"]}"
+      form.errors.messages.inspect.must_equal "{:email=>[\"is missing\"], :username=>[\"is missing\"]}"
     end
   end
 
@@ -293,7 +293,7 @@ class ValidationGroupsTest < MiniTest::Spec
     # invalid.
     it do
       form.validate({email: 9}).must_equal false
-      form.errors.messages.inspect.must_equal "{:username=>[\"username is missing\"]}"
+      form.errors.messages.inspect.must_equal "{:username=>[\"is missing\"]}"
     end
   end
 
--- a/test/validation/dry_validation_test.rb
+++ b/test/validation/dry_validation_test.rb
@@ -15,26 +15,20 @@ class ValidationGroupsTest < MiniTest::Spec
       property :confirm_password
 
       validation :default do
-        key(:username, &:filled?)
-        key(:email, &:filled?)
+        key(:username).required
+        key(:email).required
       end
 
       validation :email, if: :default do
-        key(:email) do |email|
-          email.min_size?(3)
-        end
+        key(:email).required(min_size?: 3)
       end
 
       validation :nested, if: :default do
-        key(:password) do |password|
-          password.filled? & password.min_size?(2)
-        end
+        key(:password).required(min_size?: 2)
       end
 
       validation :confirm, if: :default, after: :email do
-        key(:confirm_password) do |confirm_password|
-          confirm_password.filled? & confirm_password.min_size?(2)
-        end
+        key(:confirm_password).required(min_size?: 2)
       end
     end
 
@@ -52,7 +46,7 @@ class ValidationGroupsTest < MiniTest::Spec
     # invalid.
     it do
       form.validate({}).must_equal false
-      form.errors.messages.inspect.must_equal "{:email=>[\"email must be filled\"], :username=>[\"username must be filled\"]}"
+      form.errors.messages.inspect.must_equal "{:username=>[\"username must be filled\"], :email=>[\"email must be filled\"]}"
     end
 
     # partially invalid.
@@ -65,7 +59,7 @@ class ValidationGroupsTest < MiniTest::Spec
     it do
       form.validate(username: "Helloween", email: "yo!", confirm_password:"9").must_equal false
       form.errors.messages.inspect
-        .must_equal "{:confirm_password=>[\"confirm_password size cannot be less than 2\"], :password=>[\"password must be filled\", \"password size cannot be less than 2\"]}"
+      .must_equal "{:confirm_password=>[\"confirm_password size cannot be less than 2\"], :password=>[\"password must be filled\", \"password size cannot be less than 2\"]}"
     end
     # 4th group with after: fails.
     it do
@@ -73,7 +67,6 @@ class ValidationGroupsTest < MiniTest::Spec
       form.errors.messages.inspect.must_equal "{:confirm_password=>[\"confirm_password size cannot be less than 2\"], :password=>[\"password must be filled\", \"password size cannot be less than 2\"]}"
     end
   end
-
 
   describe "Nested validations" do
     class AlbumForm < Reform::Form
@@ -103,25 +96,21 @@ class ValidationGroupsTest < MiniTest::Spec
       end
 
       validation :default do
-        configure { |config|
-          # config.messages_file = 'test/fixtures/dry_error_messages.yml'
-        }
+        configure do
+          option :form
+          # message need to be defined on fixtures/dry_error_messages
+          # d-v expects you to define your custome messages on the .yml file
+          def good_musical_taste?(value)
+            value != 'Nickelback'
+          end
 
-        key(:title) do |title|
-          title.filled? & title.good_musical_taste?
+          def form_access_validation?(value)
+            form.title == 'Reform'
+          end
         end
 
-        key(:title, &:form_access_validation?)
-
-        # message need to be defined on fixtures/dry_error_messages
-        # d-v expects you to define your custome messages on the .yml file
-        def good_musical_taste?(value)
-          value != 'Nickelback'
-        end
-
-        def form_access_validation?(value)
-          form.title == 'Reform'
-        end
+        key(:title).required(:good_musical_taste?)
+        key(:title).required(:form_access_validation?)
       end
     end
 
@@ -142,9 +131,9 @@ class ValidationGroupsTest < MiniTest::Spec
       result = form.validate(
         "title"  => "Reform",
         "songs"  => [
-                      {"title" => "Fallout"},
-                      {"title" => "Roxanne", "composer" => {"name" => "Sting"}}
-                    ],
+          {"title" => "Fallout"},
+          {"title" => "Roxanne", "composer" => {"name" => "Sting"}}
+        ],
         "band"   => {"label" => {"name" => "Epitaph"}},
       )
 
@@ -238,11 +227,11 @@ class ValidationGroupsTest < MiniTest::Spec
       property :email
 
       validation :email do
-        key(:email, &:filled?)
+        key(:email).required
       end
 
       validation :email, inherit: true do # extends the above.
-        key(:username, &:filled?)
+        key(:username).required
       end
     end
 
@@ -256,7 +245,7 @@ class ValidationGroupsTest < MiniTest::Spec
     # invalid.
     it do
       form.validate({}).must_equal false
-      form.errors.messages.inspect.must_equal "{:username=>[\"username must be filled\"], :email=>[\"email must be filled\"]}"
+      form.errors.messages.inspect.must_equal "{:email=>[\"email must be filled\"], :username=>[\"username must be filled\"]}"
     end
   end
 
@@ -270,20 +259,17 @@ class ValidationGroupsTest < MiniTest::Spec
       property :password
 
       validation :email do
-        # validates :email, presence: true
-        key(:email, &:filled?)
+        key(:email).required
       end
 
       # run this is :email group is true.
       validation :after_email, if: lambda { |results| results[:email]==true } do # extends the above.
-        # validates :username, presence: true
-        key(:username, &:filled?)
+        key(:username).required
       end
 
       # block gets evaled in form instance context.
       validation :password, if: lambda { |results| email == "john@trb.org" } do
-        # validates :password, presence: true
-        key(:password, &:filled?)
+        key(:password).required
       end
     end
 
--- a/test/validation/dry_validation_test.rb
+++ b/test/validation/dry_validation_test.rb
@@ -69,6 +69,8 @@ class ValidationGroupsTest < MiniTest::Spec
   end
 
   describe "Nested validations" do
+    Guitar = -> { "Purple"  }
+
     class AlbumForm < Reform::Form
       include Reform::Form::Dry::Validations
 
@@ -95,7 +97,7 @@ class ValidationGroupsTest < MiniTest::Spec
         end
       end
 
-      validation :default do
+      validation :default, with: { title: -> { title }, guitar: Guitar } do
 
         key(:title).required
 
@@ -107,20 +109,21 @@ class ValidationGroupsTest < MiniTest::Spec
         end
 
         configure do
-          option :form
+          option :title
+          option :guitar
           # message need to be defined on fixtures/dry_error_messages
           # d-v expects you to define your custome messages on the .yml file
           def good_musical_taste?(value)
             value != 'Nickelback'
           end
 
-          def form_access_validation?(value)
-            form.title == 'Reform'
+          def validation_with_dependencies?(value)
+            title == 'Reform' && guitar == 'Purple'
           end
         end
 
         key(:title).required(:good_musical_taste?)
-        key(:title).required(:form_access_validation?)
+        key(:title).required(:validation_with_dependencies?)
       end
     end
 
--- a/Gemfile
+++ b/Gemfile
@@ -13,5 +13,4 @@ gemspec
 # gem "declarative", path: "../declarative"
 
 gem "minitest-line"
-gem 'dry-validation'
 gem 'byebug'
--- a/lib/reform/form/active_model/validations.rb
+++ b/lib/reform/form/active_model/validations.rb
@@ -51,7 +51,7 @@ module Reform::Form::ActiveModel
       extend Uber::Delegates
       delegates :@validations, :validates, :validate, :validates_with, :validate_with
 
-      def call(form) # FIXME.
+      def call(form, options) # FIXME.
         validator = @validations.new(form)
         validator.valid?
 
--- a/lib/reform/form/active_model/validations.rb
+++ b/lib/reform/form/active_model/validations.rb
@@ -51,12 +51,12 @@ module Reform::Form::ActiveModel
       extend Uber::Delegates
       delegates :@validations, :validates, :validate, :validates_with, :validate_with
 
-      def call(fields, errors, form) # FIXME.
+      def call(form) # FIXME.
         validator = @validations.new(form)
         validator.valid?
 
         validator.errors.each do |name, error| # TODO: handle with proper merge, or something. validator.errors is ALWAYS AM::Errors.
-          errors.add(name, error)
+          form.errors.add(name, error)
         end
       end
     end
--- a/lib/reform/form/dry.rb
+++ b/lib/reform/form/dry.rb
@@ -30,10 +30,9 @@ module Reform::Form::Dry
 
       def call(fields, reform_errors, form)
         validator = @validator.new(form)
-
-        # a message item looks like: {:confirm_password=>[["confirm_password size cannot be less than 2"], "9"]}
+        # a message item looks like: {:confirm_password=>["confirm_password size cannot be less than 2"]}
         validator.call(fields).messages.each do |field, dry_error|
-          dry_error[0].each do |attr_error|
+          dry_error.each do |attr_error|
             reform_errors.add(field, attr_error)
           end
         end
--- a/lib/reform/form/dry.rb
+++ b/lib/reform/form/dry.rb
@@ -29,11 +29,12 @@ module Reform::Form::Dry
         @validator = Builder.new(@schemas.dup).validation_graph
       end
 
-      def call(fields, reform_errors, form)
-        # a message item looks like: {:confirm_password=>["confirm_password size cannot be less than 2"]}
-        @validator.with(form: form).call(fields).messages.each do |field, dry_error|
+      def call(form)
+        # a message item looks like: {:confirm_password=>["size cannot be less than 2"]}
+        @validator.with(form: form).call(form.to_nested_hash).messages
+        .each do |field, dry_error|
           dry_error.each do |attr_error|
-            reform_errors.add(field, attr_error)
+            form.errors.add(field, attr_error)
           end
         end
       end
--- a/lib/reform/form/dry.rb
+++ b/lib/reform/form/dry.rb
@@ -21,32 +21,42 @@ module Reform::Form::Dry
 
     class Group
       def initialize
-        @validator = Class.new(ValidatorSchema)
+        @schemas = []
       end
 
       def instance_exec(&block)
-        @validator.class_eval(&block)
+        @schemas << block
+        @validator = Builder.new(@schemas.dup).validation_graph
       end
 
       def call(fields, reform_errors, form)
-        validator = @validator.new(form)
         # a message item looks like: {:confirm_password=>["confirm_password size cannot be less than 2"]}
-        validator.call(fields).messages.each do |field, dry_error|
+        @validator.with(form: form).call(fields).messages.each do |field, dry_error|
           dry_error.each do |attr_error|
             reform_errors.add(field, attr_error)
           end
         end
       end
-    end
 
-    class ValidatorSchema < Dry::Validation::Schema::Form
-      def initialize(form)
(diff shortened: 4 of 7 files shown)


blelump:
The build fails due to missing dependencies of `dry-{validation, types, logic}` gems, as I provided them only in `gemfiles/Gemfile.rails-{3.1,4.2}`. For these build [passes](https://travis-ci.org/apotonick/reform/builds/115105220), so I don't bother about the rest. Let's wait until @solnic releases new `dry-v` and I'll bump to new version in gemspec.

blelump:
@apotonick , I've spoken with @solnic on gitter and he suggests to change the way `form` instance [is currently passed](https://github.com/blelump/reform/commit/de6769ee4c9f6ee5188b18230694daf5449cb9c7#diff-1a4da032b10542c1192263712a2d982cR34) through the new DSL, because every time it's set (and now it's set every time literally ;-), the rules have to be recompiled.

I've noticed that the `form` dependency exists from the beginning of the `dry.rb` file existence, introduced by @mrbongiolo . I've personally called this dependency a few times to get the model, but didn't get into any complex use case. If we would want to be flexible and let user set validation context per validation (there're edge cases for sure for such scenario), then probably `validation` method has to receive validation content, e.g:
```
validate(:default, with: {smth: model}) do
  key(:email).required
end
```

It seems API is almost ready for such approach. See [here](https://github.com/blelump/reform/blob/master/lib/reform/validation.rb#L10)

ps. before saying anything, [see](https://github.com/dryrb/dry-validation/blob/master/spec/integration/custom_predicates_spec.rb#L68) how it works now in `dry-v` :-)

mrbongiolo:
@blelump +1. We just had the `form` instance at the time because some scenarios needed access to the model or the form values. But just passing it whenever it's really needed seems as a better solution. Not sure if `dry-v` already has access to the "object" values, I remember at the time you couldn't do some validations without customizing them completely.

solnic:
> I remember at the time you couldn't do some validations without customizing them completely.

@mrbongiolo do you remember any specific examples?

solnic:
@mrbongiolo making a schema depend on a whole form just to do db-uniqueness validation is a code smell. I'll be adding a new interface for injecting dependencies from "the outside" so that you can have your model injected (or whatever that's needed for db interactions, ie a rom relation) when setting up a schema. With current approach I bet reform+dry-v is super slow :/

mrbongiolo:
@solnic it surely is, but that was the easiest way to get it going at that time, we can always improve it. That's why I vote for the simplest solution, to just have the basic schema there as you would using bare dry-v, and just add more stuff to it if people start having too many edgy cases. But that's up to @apotonick also.
Injecting dependencies on the Schema setup seems fair enough, or even as @blelump suggested, have the reform validation block accept some options that I could pass down to the dry-v schema.
In pure dry-v it would be quite easy, since (if I'm not wrong) I can just create a custom method call whatever I need and just return the expected true/false. Sadly this gets a bit messy inside the validation block :/

solnic:
You can now do this:

``` ruby
schema = Dry::Validation.Schema do
  configure do
    option :model, User

    def unique_email?(value)
      User.where(email: value).exists?
    end
  end

  key(:email).required(:unique_email?)
end
```

blelump:
Well, I thought about it for a while and I'd propose to slightly change the approach in case of `dry-v` validation. Obviously, we still could support the `validation` blocks, however as I suppose, the plan is to replace `AM:V` with `dry-v` sooner or later. Besides I think the DSL has changed significantly since 0.6 and actually everything what Reform supports natively, can be done in `dry-v`.

Version 0.7 extends schema definition so such approach is valid:
```
base_form = Dry::Validation.Form do
  key(:password).required
end

schema = Dry::Validation.Schema(base_form) do
  key(:email).required
end

p schema.({}).messages
# => {:email=>["is missing"], :password=>["is missing"]}
```

Another important thing is conditional validation, e.g:
```
form = Dry::Validation.Form do
  key(:password).required
  key(:email).required

  rule(password: [:email, :password]) do |email, password|
    # Validate password if email is valid
    email.email_valid?.then(password.valid?)
  end

  configure do
    def self.messages
      Dry::Validation::Messages.default.merge(
        en: { errors: { valid?: 'must be a valid ' }  }
      )
    end

    def email_valid?(email)
      email.include?('@')
    end

    def valid?(pass)
      false
    end
  end
end
p form.('email' => '5', 'password' => '123').messages
 # => {}
```

Essentially what I'd propose is not to wrap `dry-v` logic with validation blocks and provide new method that will handle pure `dry-rb` Schema object, e.g:
```
validates_with Dry::Validation.Schema {}
```

The `validates_with` could register `dry-rb` Schema objects and then just call validation with form data and deliver errors. If you guys think about any edge use case that wouldn't fit, please provide it here so we could investigate it.

mrbongiolo:
Where would you define the validations Schema?

blelump:
@mrbongiolo  , we'd have at least 2 solutions:
```
validates_with Dry::Validation.Schema do
  key(:email).required
end

# or:

validates_with User::Schema::Validation
# user/schema.rb
module User::Schema
  Validation = Dry::Validation.Schema do
    key(:email).required
  end
end

```

ps.  I've used `validates_with` accidentally, because now I see it's already defined. I wouldn't touch too much the AM:V related API and use some custom method.

solnic:
I agree with @blelump. When using dry-v you should be able to do everything that reform validation "bonuses" provide, and if it's not the case I'll be happy to cover any missing features.

BTW one small nitpick - I don't like how the term "conditional validation" is being used. I understand where it's coming from but we should really think about validation logic as something that's *always conditional*. In example, you don't want to validate a value under a key when that key is not present, or you don't want to validate a nested hash when it's not really a hash, etc. This is something that's very explicit in dry-v, you can easily see it when you look at validation rule ast (assuming you have a big screen :laughing:). This idea is one of the core aspects of dry-v, right next to type-safety.

mrbongiolo:
@blelump maybe we could stay with the `validation` but it could directly define `Dry::Validation.Schema` in the way that you suggested. After all the reform validation block is a concept to hold validation. Just not sure how this would play with AM:V.
@solnic AFAIK reform's validation does nothing special, actually it provides some conditional and group validations, but this is easily done when using dry-v instead. 

What about we have the `validation` to work with dry-v, and let AM:V as it's, using `validates :blargh`?

mrbongiolo:
TBH I would remove all group and conditional stuff from `validation`. That's not something that reform should really care about, I guess it came to be because AM:V doesn't have a good way to handle it. But it would be better overall for reform to just support basic AM:V `validates :something` and use validation as a dry-v Schema. But that requires @apotonick feedback.
I guess this could simplify a good amount of things, and in the future have the majority of users using dry-v.

blelump:
@mrbongiolo , I'm fine with `validation`,  `validates` etc, however at first, I'd move all that AM:V related logic from  [`Reform::Validation`](https://github.com/apotonick/reform/blob/master/lib/reform/validation.rb) module. It's really coupled with AM:V, or to be more general, it's all designed to support block contexts, that fire somewhere else. I think, that with new `dry-v` DSL we don't need it at all, except a small piece of code, that will register `Schema` objects, perform validation against form data and track errors. I'd also move `Reform::Validation::Groups`, because it's related as well. Perhaps `reform_am_v` gem would be a good idea. 

It's worth noting that if we would prepare such simple logic, it will work with any object that responds to `call` and `messages`. I mean, one would provide an arbitrary type, e.g:
```
class Validator
  def self.call(params)
    Result.new
  end

  class Result
    def messages
      {}
    end
  end
end

p Validator.call({}).messages
 # => {}
```

Isn't it cool? :-)

apotonick:
I knew that dry-will provide features like "conditional" validations at some point, but until today, that wasn't available so I designed a generic interface which allows this pretty cool feature with AM:V, too.

Also, I am more than happy to remove things from Reform in favour of dry-v, but this will cause pain for people still using AM:V. We should find a nice transitional solution, even though I personally am *zero* interested in supporting anything Rails anymore (it is a waste of time). :grimacing: 

As for the `form` dependency - we should provide that per default using the new API, this is very commonly needed (at least that's what I remember). Will that still cause a recompilation of the rules? Excuse my ignorance, but any summarising pointer here highly appreciated!

apotonick:
@blelump Thanks so much for working on this, BTW! I am pretty confident that with yours and @solnic's help we can push a new Reform/Dry-v integration pretty easily.

Maybe we could really "deprecate" (move to separate gem) the validation groups, etc. since Dry-v will do this for us. People can then smoothly learn the new API without any real trouble.

mrbongiolo:
@apotonick would you be up to have only dry-v as default option for Reform? And then the separate reform-am-v gem that could be used to add basic support for AM:V. I don't think we should support validation blocks for AM:V, if people really need conditional validation they can either hack it with `if:` or just use dry-v.

solnic:
> As for the form dependency - we should provide that per default using the new API, this is very commonly needed (at least that's what I remember).

Why is that? A schema, typically, should receive its external deps (if any are needed) during first initialization. There should be no need to inject form object into schema. In fact - schema should be called prior creating a form object...dry-v's philosophy is to validate (and potentially sanitize and/or coerce) input prior creating *any objects using that input as the state*.

> Will that still cause a recompilation of the rules? Excuse my ignorance, but any summarising pointer here highly appreciated!

Rules are recompiled every time you do `schema.with(...)` to inject additional objects into a schema.
Rules are recompiled every time you do `schema.with(form: form)` so if you need to inject a new form object every time you validate, then yes.

blelump:
Hi guys !

I think I'll handle it, however today I've got into an issue, that unveiled how this logic is complicated :smile: . For instance, look at this backtrace jumps:
```
reform/lib/reform/contract/validate.rb:18:in `validate!'
reform/lib/reform/contract/validate.rb:30:in `block (2 levels) in validate_nested!'
disposable-0.2.5/lib/disposable/twin/property_processor.rb:29:in `property!'
disposable-0.2.5/lib/disposable/twin/property_processor.rb:17:in `call'
reform/lib/reform/contract/validate.rb:30:in `block in validate_nested!'
disposable-0.2.5/lib/disposable/twin/definitions.rb:22:in `block in each'
declarative-0.0.5/lib/declarative/definitions.rb:30:in `each'
declarative-0.0.5/lib/declarative/definitions.rb:30:in `each'
disposable-0.2.5/lib/disposable/twin/definitions.rb:16:in `each'
reform/lib/reform/contract/validate.rb:28:in `validate_nested!'
reform/lib/reform/contract/validate.rb:16:in `validate!'
reform/lib/reform/contract/validate.rb:30:in `block (2 levels) in validate_nested!'
disposable-0.2.5/lib/disposable/twin/property_processor.rb:29:in `property!'
disposable-0.2.5/lib/disposable/twin/property_processor.rb:17:in `call'
reform/lib/reform/contract/validate.rb:30:in `block in validate_nested!'
disposable-0.2.5/lib/disposable/twin/definitions.rb:22:in `block in each'
declarative-0.0.5/lib/declarative/definitions.rb:30:in `each'
declarative-0.0.5/lib/declarative/definitions.rb:30:in `each'
disposable-0.2.5/lib/disposable/twin/definitions.rb:16:in `each'
reform/lib/reform/contract/validate.rb:28:in `validate_nested!'
reform/lib/reform/contract/validate.rb:16:in `validate!'
reform/lib/reform/contract/validate.rb:10:in `validate'
reform/lib/reform/form/validate.rb:23:in `validate'
trailblazer-1.1.0/lib/trailblazer/operation.rb:144:in `validate_contract'
```
This is the output from `caller` called from withi `valid?` method (`Reform::Validation`). The stacktrace gives me information that all the twins are populated way before the validation performs and I believe it should be the opposite. First, we perform validation and then Twin(s) population appears (see also summary of this post). 

Essentially I've got into this issue because of nested data, e.g:
```
property :student do
    property :user do
      property :email
    end
  end
```
and it gets deserialized into `{"phone_number"=>"+48 232343242234", "foreign_student"=>#<#<Class:0x000000039aae90>:0x00000004359b3...` so it's not a params hash. I recall this, because [here](https://github.com/apotonick/reform/blob/master/lib/reform/validation.rb#L36) you explictily pass `@fields` (got in real troubles while trying to find where it's defined and finally used `ObjectSpace` to localize it – hello disposable :smile:  ) and hence, `dry-v` crashes  for nested schemas. Besides, we do not have any spec scenario for that  :smile:  . So I've used `to_nested_hash` and replaced the `@fields` variable. https://github.com/blelump/reform/commit/3f6744ce23d873c4573bcaf8ff4df2b789f3bfa5#diff-6eacdc20d4fdded3689e3cf714969cbfR36

To sum up, I think before considering the topic of this PR, we need to deal with Reform internals and probably move Twin population after validation. Do you see any potential issues with this? I know that there're use cases where data validation is simply not enough, but then I believe, if user would need either a twin or a model, he can instantiate that on demand. Besides, how such changes are related to AM:V?

mrbongiolo:
If I'm not wrong reform uses twins and disposable to coerce the values. This probably could be done with dry-v or one of the other dry-rb gems. But this would require some reworking on other core parts of Reform. 

Right now nested hashes doesn't play nicely between reform and dry-v because each nesting is basically a new reform form object. Also reform still uses Virtus for coercion, which @solnic himself is not supporting anymore. We could try to use dry-values here also. 

Or leave everything else as it is and have dry-v just do validation, even in a not so good manner.

blelump:
@mrbongiolo , yep, inferring type coercions is provided by `Disposable::Twin` and Virtus internally (currently it has a successor called `dry-types`), however I wouldn't touch it, at least for now. 

Right now we should focus on `dry-v` and how to provide all of its features with minimal impact. Besides, we need to find out how to deliver `dry-v`logic that allows one to inject an arbitrary dependency, e.g. having such schema:
```
validation Dry::Validation.Schema do
        key(:email).required(:email?)

        configure do
          option :email_check

          def email?(value)
            email_check.(value)
          end
        end
end
```
how to inject `email_check`. [Here's](https://github.com/dry-rb/dry-validation/blob/7cf87849bf7e5b001d0f34efb7193a368f135b5e/spec/integration/custom_predicates_spec.rb#L70) a complete example.

blelump:
Hi guys,

@solnic just released `dry-v` 0.7 :clap: , so we can start dealing with it officially.

apotonick:
@solnic I agree with you, but in real life, validations in a "Rails" application (let's call it "MVC") will often need dependencies, e.g. the current user, which is why I like your injection mechanism.

In my understanding, a validation verifies a deserialised object graph. In your philosophy it seems to be more of a "formal" validator, is that about right? The cool thing is our two different views on that still work together because dry-v kicks ass. :wink:

blelump:
@apotonick , @mrbongiolo ,

the more I'm using this, the more issues appear :disappointed: . In particular:

* I cannot use `dry-v` coercions, because the validation result is not populated in twins.
* I cannot use `virtus` coercions either, because it infers coercion right before `dry-v` performs and therefore `dry-v` complains about smth is not a `String` type. Quite obvious, but annoying...

My proposal is to re–orchestrate Reform so in the end:

1. `dry-v` validation is performed first – no twins exist at this state.
2. Twins are populated with the output data from `dry-v`. If one used `dry-v` coercions, twins are populated with appropriate types. Obviously, twins might be populated with invalid data, so that one is able to re–render the form with errors.

More TODOs:
* improve `dry-v` error parsing, because now schema might contain another schema ([and another]), which in the end result with output smth like `{a: {b: ['is missing']}}`. This is currently not covered.
* keep support for AM:V and move AM:V related logic somewhere (new gem would be a best fit?)

Do you guys have any caveats with that? In case of re–orchestration there's a lot work to do so any help is appreciated.

apotonick:
@blelump "Re-orchestrating" basically here means changing the way Reform works fundamentally, which I do not second. In my understanding, the process is

1. Deserialize the incoming document into an object graph.
2. Coerce incoming values according to the form's settings.
3. Validate the deserialized object (form) graph.

This workflow has many advantages, e.g. you can have "business validations" like "user is authenticated and can't have more than 5 ocmments, where the last comment must have been added a week or more ago" in your contract.

The validating object graph is a concept I've worked on a lot (and reflected a lot) and is something I'd rather not give up because it's incredibly nice to work with. 

We should rather separate coercion and validation, unless there's a good reason to mix those. I was actually working on that a while ago, I was using `dry-types` coercing. Objections? Ideas?

blelump:
Hi @apotonick , thanks for response.

I get your point of view, so let's think about it. If we'd like to validate the deserialized object (I assume that's just twin objects basically) and some of its params have already been coerced (with `virtus` or `dry-types`), then with `dry-v` we'd need to use a `Schema` instead of a `Form`. Am I right @solnic ? I assume with `Schema` it'll work with all primitives. So the process might look like:

1. Deserialize the incoming document into an object graph.
2. Coerce incoming values according to the form's settings.
3. Serialize object graph into hash of primitives or:

  __Caveat__: @solnic , would it be possible that `dry-v`  eats object and checks rules against object instead of a hash? In such case, I would be able to pass e.g an `OpenStruct` object, e.g. `schema.(OpenStruct.new({name: [1]}))`. That would significantly simplify Reform validation process. What do you think?
4. Validate the (de)serialized (in parenthesis because it depends upon @solnic response of the previous point) object graph. 

Besides, I'm still thinking about the current approach (using blocks), as the new `dry-v` DSL is already implemented within this PR and we do have a plan to handle coercions as a separate concern.

mrbongiolo:
@blelump So as @apotonick said he doesn't want to change the way reform deserialize and coerce the incoming hash. This way we need `dry-v` to validate those values only. For simple non-nested hashes it's fine, just put the schema there and you can even use coercion rules, but always knowing reform had already coerced the hash.

Where it could get messy is with nested-hashes, since a reform form is initialized for each nested `property` block.

One option is to have a single Schema validation block for an entire form, and send to it a complete hash like: `{title:'foo', artist: { full_name: 'bar' } }`.
```ruby
class AlbumForm < Reform::Form
  property :title

  property :artist do
    property :full_name
  end

  # maybe the rules are wrong, but you get the idea
  schema_validation do
    key(:title).required

    key(:artist).required do
      key(:full_name).required
    end
  end
end
```

Other is to have the schema only access the "current" form level, something like this:
```ruby
class AlbumForm < Reform::Form
  property :title

  property :artist do
    property :full_name

    schema_validation do
      key(:full_name).required
    end
  end

  schema_validation do
    key(:title).required
  end
end
```

I prefer the first version, but this would mean some heavy changes on the way Reform operates, so what we can do here is have a basic validation integration. Maybe I'm wrong and miss something here, but that's the way I see it for the current state of reform.

@blelump maybe a dry-rb reform version could be an option, you keep some of the good stuffs from reform but could fully integrate `dry-types`, `dry-v` and such. But then this is another huge change.

mrbongiolo:
@apotonick Would it be possible to drop direct support for AM:V and just use dry-v? I know that the majority of users still are on AM:V, but dry-v is  a much better option.

Bump the version and people who really wants AM:V can stay on this version, or we can try to use a reform-rails gem. But even with such a gem reform itself need to know a lot about how dry-v and AM:V work, since in dry-v we can basically just `call` and then get the messages, but AM:V has to be "injected" in the form class directly, and then nested forms have their own errors that the parent form need to know about...

solnic:
Main idea behind dry-validation is to validate/coerce/sanitize untrusted data prior using them to initialize an object. In that sense, initializing reform form before validating is against that philosophy.

One of the great features of dry-validation is treating type-safety as a first class concept. As I understand, you can use reform's form to perform more work on the attributes, like setting some values, coercing some values and maybe even more. This is something that depends on the validity of the input. If it's not valid, you shouldn't touch it. This is why applying schema prior digesting data by reform's form would be a better design decision. In fact, I believe it would allow you to simplify reform's implementation as you wouldn't have to worry about validity of the data, as you're gonna be sure it's valid and properly coerced already.

Simple example - how do you coerce an empty string? Based on what you're gonna make the decision what to do with it? If this happens prior validation, then you can't really tell what the expectation is when it comes to a value that is empty. This means that coercion that is detached from validation simply cannot be correct. You can provide some default behaviors and hope it's gonna be correct in most of the cases but it will never-ever be 100% correct. This is something dry-v *solved*.

It took me a while (for whatever reason!) to understand how coercion should work, I knew for a long time that it's strictly related to validation, but I couldn't figure out how to approach it until now. And it's pretty simple: coercion should be based on schema definition.

Now, having said that - it's not like it means you can't or shouldn't use dry-validation in reform. You can, and it's gonna be better than AM::Validation. It's just that you can't leverage its complete features wrt coercion and you're gonna have an API where you have duplicated definitions of the same thing, because first you define form's properties, then you define validation schema. Both define the same structure, that's gonna be a problem I believe, because people hate boilerplate.

blelump:
Hi @apotonick , what's your feelings about what @solnic said?

apotonick:
Very helpful discussion! :heart: 

I fully understand now that dry-v's philosophy is to "mix" coercion and validation, and I like that, since they can't be fully separated. What I'm still struggling to understand, though, is how deserialization works in that environment. 

One of the cool concepts in Reform is that nested fragments can be deserialized to nested form objects, where the parent form doesn't know anything about that process. This can be very complex, and that's because deserializing documents *is* complex, and Reform solves this very elegantly with its populators. Unfortunately, deserialization/population happens before any validation takes place (so far). I can't really see how dry-v tackles that, you always speak about nested hashes, but deserialization is a bit more than just parsing a hash - please help me understanding your thoughts here.

Generally, I see two options we have.

1. The first example that @mrbongiolo posted, where each form uses dry-v for validation, coercion still happens at deserialization (using dry-t). That's basically what we have right now.
2. Apply dry-v before we deserialize to the incoming hash, then only let Reform do the deserialization part (since coercion and validation already happened). I am not sure what are the benefits here.

I am also open for allowing both, and see what works best for us. Long term, I want to remove as much code as possible (e.g. the groups and conditionals), put that into `reform-activebullshit` and use dry-v wherever possible.

solnic:
> Very helpful discussion! :heart:

It is <3

> I fully understand now that dry-v's philosophy is to "mix" coercion and validation, and I like that, since they can't be fully separated.

It "mixes" these two things but these are two distinct steps. We apply input-processing to the input prior applying the rules. And it's optional behavior. You can have a plain sanitizer that will simply clean your input from unexpected keys, or a sanitizer+coercer that will also try to coerce strings into other types.

> What I'm still struggling to understand, though, is how deserialization works in that environment.

Please define `deseralization` first, and I'll tell you.

> I can't really see how dry-v tackles that, you always speak about nested hashes, but deserialization is a bit more than just parsing a hash - please help me understanding your thoughts here.

dry-v cares about the data structure and the values. It doesn't decorate the data with additional behaviors. So if `deserialization` means building objects from the input, then dry-v doesn't have (and won't have) any interfaces for doing it.

> 1. The first example that @mrbongiolo posted, where each form uses dry-v for validation, coercion still happens at deserialization (using dry-t). That's basically what we have right now.

I need to fully understand what deserialization means to say more :)

> 2. Apply dry-v before we deserialize to the incoming hash, then only let Reform do the deserialization part (since coercion and validation already happened). I am not sure what are the benefits here.

There's a huge benefit here: you don't have to worry about the validity of the input, because if schema result is successful then you know everything's fine. Otherwise your reform objects need to process data that are potentially broken, but then again - I don't really understand what deserialization involves :)

blelump:
@solnic , deserialization is a process of buliding the object graph based on the input data. In particular, it creates a twin objects which are the 'twin' representation of real objects with the data you've provided. 

Right now the deserialization process always happens first. If the incoming data is invalid, object graph (twins representation) is built upon invalid data and validated afterward, using whatever validation library.

solnic:
@blelump what's the responsibility of twins?

blelump:
@solnic ,

the "twin" representation exists in Reform, because of these things:
* The underlying Reform validation feature bases on AM:V. As we know, AM:V validates the objects itself so the  purpose of the 'twin' representation is to validate possibly invalid objects against defined AM:V rules. If such representation is valid, you can persist its state to the original models (by simply calling `twin.save`). The twin idea is implemented in the [disposable](https://github.com/apotonick/disposable) gem.
* If validation fails, you're able to present invalid form, e.g. when provided data does not match validation schema. Most of the "popular" form renderers from the Rails world present an object graph ('twin" representation in Reform case). It would be diffcult to do that without such representation. 

  **note**:
  even having `dry-v` validation step as the first step of the process, you'd need to create such twin representation to let one render invalid form. Another (but similar?) approach is presented in `formalist` gem, but I am not sure (at least by now), how one would persist the form data with the formalist approach (including collections, relations etc) without the deserialization process (in Reform case, deserialization creates twins of models).

mrbongiolo:
@apotonick can you accept this PR? It doesn't fix all things that we're discussing, but it adds the necessary support for `dry-v` 0.7+

After that we can work on another release that changes the way they both interact and hopefully remove and discard more `active-stuff`.

blelump:
@mrbongiolo , after a while, I personally don't feel comfortable with this PR (at least in the current form) :-) , even though it introduces 0.7 support. 

**First let me clarify where are we now:**
* This PR would introduce even more mess, because despite the new DSL works (partially), it completely doesn't address coercion problem. Moreover, `virtus` coercions will not work too, because we use `DryV::Form` which explictly expects string primitives. We could change it to a `DryV::Schema`, but then all keys need to be coerced into symbols. We could ask @solnic about another input sanitizer, but I feel this is just wrong, because we try to adjust something, that doesn't fit, because of how it's been designed.
* Even though we would make the validation process first (before deserialization), it still needs to create invalid objects after validation (no matter if succeeded or failed), because most of the form renderer libraries expect object tree to render invalid form.

------


**What can be done with minimum effort:** `# it also covers validate each form as a separate hash of primitives`

_Summary:_
Minimum effort assumes that the whole process stays the same, which means that deserialization happens first. We also do not abandon the `validation(&block)` approach, because for Reform design it fits better.

Things to complete before switching to 0.7:
*  Ensure coercions work with `dry-t` (note I don't mention `dry-v`). 
  TODO: 
  * introduce `dry-t` and get rid of `virtus` for coercions.

  * ensure `dry-v` eats hash of primitives, even for complex object graphs (achievable with `Twin#to_nested_hash` method). 
  Comment: 
  You may think that's stupid, because first we deserialize and then serialize again, but still I vote for it, because as an alternative we've got validation of each nested structure as a separate invocation, which I discourage. Besides, it's more consistent to have validation schema in one place and for one object graph.



* Improve error handling for nested schemas. Right now it works only for a flat schema (without any nested data).
  TODO:
  * parse error output and adjust to Reform needs.

What we get with such approach:
* `dry-v` gets a hash of primitives (already coerced) and validates them against schema rules. As a result of `dry-v` job, we get only validation result (status + errors) without any coercions;
* _`dry-v` schema is fully supported_ (**!**), including `with` option, which might be helpful for users to inject extra deps;
* _cons_: we introduce extra boilerplate, because coercions have to be defined in a twin schema, e.g: `property :id, type: Types::Strict::Int`


We get almost fully featured `dry-v` (without schema inheritance). Would you guys agree? // cc: @apotonick , @mrbongiolo , @solnic (?)

mrbongiolo:
I was going to write something on why we could stick with flat schema and such for `reform-v2.1.x` and `dry-validation-v0.7.x`, but the changes required to better integrate both of them doesn't seem so hard and are good enough for a version bump on `reform`.

AFAIK coercion happens on `disposable`, so we should tackle there the use of `dry-t` instead of `virtus`.

What happens if a coercion fails when populating the twins? Should it go to the errors object?

If you only validate the data after creating the twins you can't enforce Strict Types, or am I wrong here?

------

On Reform itself we need to change the way on how all the validation occur and how it call each integration style (AM:V or dry-v). Reform already has a basic errors object, but we need to improve it a bit (it's required by those rails form rendering libs) so that Reform::Dry or Reform::AMV will in the end parse the errors from each validation library to Reform::Errors (not that hard to do).

This way we can simplify `validation(&block)`, it should perform any group stuff, or anything related to that, it would just hold the form validations, dry-v or AM:V style, the user choose.

When using with `dry-v` we would have a single `validation(&block)` for the entire form (QUESTION: does this work for collections?).

When using with `dry-v` we would have one `validation(&block)` per nested `property`, it's ugly, but AM:V doesn't work other way.

I would also propose to remove all AM:V related sugar code from the Form itself and only set your validations inside the `validation(&block)`. This way we can remove the `:validate` option from the `property` also. It's quite nice to have those things there, but the mess they create behind is not worth it. Better have everyone defining validations in a single point: `validation(&block)`.

The entire Reform form would be simplified, as it doesn't have to worry about dumb AM:V messing around it scope.

blelump:
@mrbongiolo , let me clarify.

> AFAIK coercion happens on disposable, so we should tackle there the use of dry-t instead of virtus.

:+1: 

> What happens if a coercion fails when populating the twins? Should it go to the errors object?

Nah, that would be a mess. We just need to infer a non strict coercions.

> If you only validate the data after creating the twins you can't enforce Strict Types, or am I wrong here?

Yes, you're completely right!

> When using with dry-v we would have a single validation(&block) for the entire form (QUESTION: does this work for collections?).

Yes and this is the way I'd like to handle it. I mean, the validation schema is consistent and defined in one place, instead of being scattered in every place of the twin schema.

> When using with dry-v we would have one validation(&block) per nested property, it's ugly, but AM:V doesn't work other way.

We're able to have just one block for the entire form, by leveraging `dry-v` features, e.g:
```
params = {user: {name: "Bob", pets: [{name: :cat}, {name: :tiger}]}}


validator = DryV.Schema do
  key(:name).required
  key(:pets).each do
    key(:name).required
  end
end
validator.call(params[:user])
```

mrbongiolo:
@blelump I sure agree to separate those tasks.

@apotonick said on gitter chat if we could make `dry-v` work both ways, with a single schema or one per form. It shouldn't be that hard.

I already sent a PR migrating disposable to `dry-types`. Now we need a bit of a rework on Reform::Errors, so that `dry-v` can populate the nested errors correctly. Even using nested validations on a single schema, in the end you need to populate each form's twin `@errors` instance. Maybe we would like to provide nested `@errors` directly on the parent form also, what do you think?

After that we can discuss, if needed or wanted, further integration and bigger changes on Reform.

blelump:
@mrbongiolo , in case of "flat" and "global" approach, we need one "global" error object, so we would be able to put erros from within both validations. As a result, we're expecting errors gathered from all sub–validations.

@mrbongiolo , could you try to identify which parts of the Reform code is related somewhat to AM:V and hence we would start thinking what to throw out or at least prepare a list of tasks to do?

Thanks for helping with all of this :heart:

mrbongiolo:
@blelump I'll look at it today, lets see if we can come up with a task list to get this going.

apotonick:
I merged this into https://github.com/apotonick/reform/tree/2-2 - let's get it running now!

blelump:
@apotonick , @fran-worley, @mrbongiolo , specs passes. 

Please review https://github.com/apotonick/reform/pull/347/commits/778a3572e259519eaf489a9b788ea768f820460b and drop me a line in case of any caveats.

apotonick:
@blelump Thanks, I am struggling to understand the intention of https://github.com/apotonick/reform/pull/347/commits/778a3572e259519eaf489a9b788ea768f820460b, though.

test/validation/dry_validation_test.rb
```diff
@@ -95,7 +97,7 @@ class AlbumForm < Reform::Form
         end
       end
 
-      validation :default do
+      validation :default, with: { title: -> { title }, guitar: Guitar } do
```
apotonick: Is that a dry-v feature? And where do you need this? Isn't that all nicely available via the `form` instance?

apotonick:
Ok, everyone, I've played with this and came to the following conclusions (objections welcome!):

## Nested Validations

We should allow users both styles of validations, one-for-all and per-form.

### One-for-all

Here, all validations go into one schema.

```ruby
    validation :default do
      key(:user_id).required

      key(:variants).schema do
        each do
          key(:id).required
        end
      end
    end
```

### Per-form

Here, each nested form has its own schema, kinda the way it is now.

We then have to compile errors objects for each form, so the current error API remains the same (because it is super helpful).

This will probably imply some work, but I don't see why we wouldn't allow both (or am I wrong here?).

The speed problem when having to compile the schema every validation I gladly tolerate and leave it up to our smart @solnic - I'm sure we will find a way to speed this up. This shouldn't be a blocker!

## Custom High-Level Predicates

I have one problem with dry-v, though, I couldn't figure out how to have high-level predicates that are unrelated **to any input**. For example, I want to make sure the `products` array is max. 5 items when user not logged in. We really have to find a way for those validations as they are an fundamental concept of Trailblazer and validating a deserialized object graph.

## AM::Stuff and Groups

For now, we can leave Reform's groups as they can be helpful. I find them less clumsy than the way dry-v allows "chained validations". Once this is simpler in dry-v, we can remove this feature.

Speaking of removing, I would remove all AM::V related code and throw it into `reform-active`.

test/validation/dry_validation_test.rb
blelump: @apotonick , as @solnic [said](https://github.com/apotonick/reform/pull/347#issuecomment-196900284), injecting form instance _every_ time costs us rebuilding the whole schema. Here's the [implementation](https://github.com/dry-rb/dry-validation/blob/master/lib/dry/validation/schema.rb#L179) . That's why I've prepared a logic that let's one decide, whether an arbitrary dependency is needed.

blelump:
> We then have to compile errors objects for each form, so the current error API remains the same (because it is super helpful).

> This will probably imply some work, but I don't see why we wouldn't allow both (or am I wrong here?).

Yep, I agree and that's the intention. I mean, provide errors for each declared `validation` block, so whether a `flat` or `nested` validation style is used, they both work.

> The speed problem when having to compile the schema every validation I gladly tolerate and leave it up to our smart @solnic - I'm sure we will find a way to speed this up. This shouldn't be a blocker!

Uhm, according to the whole discussion above, the only one speed issue is rebuilding schema each time, when using `with` method. All the remaining cases we cover here, is just using `plain old dry-v` features or am I wrong here?

solnic:
> I have one problem with dry-v, though, I couldn't figure out how to have high-level predicates that are unrelated to any input. For example, I want to make sure the products array is max. 5 items when user not logged in. We really have to find a way for those validations as they are a fundamental concept of Trailblazer and validating a deserialized object graph.
AM::Stuff and Groups

This is a missing feature in dry-v, specifically the ability to use a predicate which doesn't rely on any input. I'll add this in next release, so you'll be able to do:

``` ruby
rule(min_products: [:products]) do |products|
  not_logged_in?.then(products.max_size?(5))
end
```

> For now, we can leave Reform's groups as they can be helpful. I find them less clumsy than the way dry-v allows "chained validations". Once this is simpler in dry-v, we can remove this feature.

We don't have an equivalent of reform's groups. The high-level rules feature is something that could be used as the base for grouping things, I plan to add this feature under the name of "safe-points" or maybe "groups" too.

As far as clumsiness goes - there are a couple of improvements I can make to make `rule` API less verbose, I was more focused on its behavior than ease-of-use in terms of syntax. This is a very powerful feature, being able to explicitly define which values need to be valid for a given rule to be applied is *extremely useful*.

solnic:
> Uhm, according to the whole discussion above, the only one speed issue is rebuilding schema each time, when using with method. All the remaining cases we cover here, is just using plain old dry-v features or am I wrong here?

Please just ignore this for now. I will make it better somehow and you should not be blocked by this in any way.

apotonick:
Oh yeah, I love the rules feature in dry-v, @solnic, I just find it a bit verbose, but I prefer having a clear and verbose API *before* abstracting that to a nicer DSL, so, all good.

Cool, I can see light at the end of the tunnel, and how I will never ever have to use Active* again! :grimacing:

fran-worley:
Closing as a merge was finally agreed upon https://github.com/apotonick/reform/commit/cd7c3b30139ff9f5e27d85085656f30d6973664f

Not everything discussed here has been fully addressed but if these are still a problem then they should be raised as separate issues.

PR #350, opened 2016-03-17
Add a README example for #property :default

heeton:


apotonick:
Thanks Alex! I'm moving all docs over to http://trb.to, check this out: http://localhost:4000/gems/reform/api.html#disposable-api I will add links and more once the docs are completed*.

*LOL

PR #357, opened 2016-04-11
Test including setters from Module

freemanoid:
It's currently impossible to extract a setter into a module.


apotonick:
Here's how it works.
http://trailblazer.to/gems/reform/api.html#forms-in-modules
http://trailblazer.to/newsletter/2016-march.html

Issue #371, opened 2016-06-21
Getting "can't dup NilClass" with Reform 2.2.0


simonc:
Hi there,

After updating from v2.1 to v2.2 I now get the following error:
```
can't dup NilClass
…/.gem/ruby/2.3.1/gems/reform-2.2.0/lib/reform/contract.rb:33:in `dup'
…/.gem/ruby/2.3.1/gems/reform-2.2.0/lib/reform/contract.rb:33:in `block in properties'
…/.gem/ruby/2.3.1/gems/reform-2.2.0/lib/reform/contract.rb:33:in `each'
…/.gem/ruby/2.3.1/gems/reform-2.2.0/lib/reform/contract.rb:33:in `properties'
…/Projects/MyProject/app/forms/some_form.rb:2:in `<class:SomeForm>'
```

The error seems to occur because of the following code:

```ruby
def self.properties(*args)
  options = args.pop if args.last.is_a?(Hash)
  args.each { |name| property(name, options.dup) }
end
```

When no option is provided it tries to call `dup` on `nil` options

Maybe it should be replaced with

```ruby
options = args.last.is_a?(Hash) ? args.pop : {}
```

I can do a PR if no-one is already on it :smile:

apotonick:
Are you doing this? http://trailblazer.to/gems/reform/#rails-support

nbulaj:
@apotonick the same error and `reform-rails` didn't solve it.

Gemfile.lock

```ruby
    reform (2.2.0)
      disposable (>= 0.3.0, < 0.4.0)
      representable (>= 2.4.0, < 3.1.0)
      uber (~> 0.0.11)
    reform-rails (0.1.2)
      activemodel (>= 3.2)
      reform (>= 2.2.0.rc1)
```

Backtrace reference to this concept:

```ruby
class User
  class Profile < Reform::Form
    properties :username, :password
    property :password_confirmation, virtual: true

    validates :username, :password, presence: true
    validates_uniqueness_of :username, case_sensitive: false
  end
end
```

Trace:

```bash
/home/bulai/.rvm/gems/ruby-2.3.1/gems/reform-2.2.0/lib/reform/contract.rb:33:in `dup': can't dup NilClass (TypeError)
	from /home/bulai/.rvm/gems/ruby-2.3.1/gems/reform-2.2.0/lib/reform/contract.rb:33:in `block in properties'
[...]

PR #372, opened 2016-06-21
Fixing Contract.properties to handle empty options

simonc:
Previous to this commit, calling `properties` without providing
an options Hash would result in an attempt to dup `nil`.

This fixes the issues and adds basic specs for `Contract.properties`.

Relates to #371.


apotonick:
:heart_eyes:

PR #377, opened 2016-07-01
Upgrade to Dry validation v 0 8 0

fran-worley:
Upgrade reform to use dry-v 0.8.0 & benefit from many bug fixes, features and performance updates.

Note that there is something wrong with bundler + rubygems.org which doesn't pick up the new dry-logic. You need to run `gem install dry-validation` then `bundle install`

Hence why travis is broken !


fran-worley:
Can we also discuss enabling global config of dry-v schemas? We could support a setting in reform initializer that simply passes it's block to into each schema. 

Currently I have to define my custom error_messages file and collection of custom predicates in every validation group. I use the same one for everything!

apotonick:
No, @fran-worley, no global configs!

apotonick:
Just kidding, of course, an initializer and a sane default sounds great to me (I am having the same problem haha).

fran-worley:
I'll play about with this over the weekend. It looks like global configs are supported in dry-v so we might not have to wrap it at all...

PR #384, opened 2016-07-22
fix dry-messages error

fran-worley:
@apotonick this should sort your issues with 0.9.3


fran-worley:
This is included in #385 so closing

Issue #378, opened 2016-07-02
Allow global configuration for dry-validation


gorkunov:
Current dry-v version allows global configuration (useful for custom predicates and messages settings):
```ruby
Dry::Validation::Schema.configure do |config|
  config.messages = :i18n
  config.predicates = MyPredicates
end
```
Currently there is no way to handle this in reform and we need manually configure each `validation` block in form. 
It would be much useful to set this configuration only once in global reform config.

Related discussion https://github.com/apotonick/reform/pull/377

fran-worley:
As of Reform 2.3 you will be able to either monkey patch reforms base schema (https://github.com/apotonick/reform/blob/master/lib/reform/form/dry.rb#L26) or pass your own into a validation block see this test for an example:
https://github.com/apotonick/reform/blob/master/test/validation/dry_validation_test.rb#L137

siassaj:
Dry::Validation::Schema::Form.configure do |config|
  config.messages = :i18n
end

fran-worley:
That is handy to know, I assumed you had to define a base class as per their docs:
http://dry-rb.org/gems/dry-validation/basics/working-with-schemas/

siassaj:
wow. I just googled 'configure dry validations globally' and found my own answer ...

siassaj:
wow. I just googled 'configure dry validations globally' and found my own answer ...

micdahl:
@fran-worley (or someone else)
Could you please explain, how concrete to get i18n running? I have tried different methods which did not lead to a satisfying result (see [reform-rails issue](https://github.com/trailblazer/reform-rails/issues/57))

PR #385, opened 2016-07-25
Improve dry integration

fran-worley:
1) provide a reform base schema which adds `option :form`
2) add `schema_class` as an option on a validation group which will be used instead of the reform base class


fran-worley:
@apotonick I know that this can't be merged yet as it messes with the group API which will break reform-rails. Any ideas?

fran-worley:
I've drafted a PR for reform-rails to migrate the AM methods and enable passing options to Group.new. Just need to fix a pre-existing failure before I publish it

fran-worley:
Fixes #378

rosskevin:
@fran-worley I'm using a lot of schema extension, it might be nice to shorten `schema_class: Foo::Schema` to `schema: Foo::Schema` since it is likely to look like a class reference anyway.  

Also, since using the `schema` option, it seems (from reading the code) that I need to name every group. It would be nice if it the dsl would see I am using an option and not make me pass `:default` as the group name.

Please correct me if I've misunderstood any of the above.

fran-worley:
Noted on the option name. Happy to change. You shouldn't have to pass in a name to every group. It defaults to :default and this branch makes no changes there.

rosskevin:
I've just confirmed that the earlier 70013bb482fe6ab329e225aa1d7b20f5ca558d29 fails when you use `schema_class` without a group name (none of my base predicates are found).  I'll pull this latest and give it another shot.

rosskevin:
Group name is required when using `schema_class` on latest 3746bf23ec8478f8923a5e4f9ef70330c518a76c as well.

fran-worley:
@rosskevin You are quite right, now fixed. I've also renamed the option to just :schema

PR #387, opened 2016-07-29
ensure heritage uses the resolved name 

rosskevin:
(same as validation group) from hash or positional args.  This is needed to be compatible with @fran-worley's #385, but with the deprecation change from @apotonick's 04c4063a0a87073811ca2170147540570b31dfcf.  The latest commit after fran's merge allowed a hash to be passed as a name to heritage.

### DIFF
--- a/lib/reform/validation.rb
+++ b/lib/reform/validation.rb
@@ -8,11 +8,10 @@ module Reform::Validation
 
     # DSL.
     def validation(name=:default, options={}, &block)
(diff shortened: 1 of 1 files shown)


apotonick:
Thanks! :heart:

apotonick:
Off-topic (let's discuss in separate ticket): you mentioned you use a lot of inheritance with your schemas. Any chance we could see a few snippets, so I get an idea how to simplify that?

PR #388, opened 2016-08-03
Improve dry integration

fran-worley:
This enables you to either validate collection via a separate validation block within the collection or via dry-v's `each { schema { # code... } }` 

Note: 
Error messages aren't being generated correctly, need to check what the messages hash should look like in AMV first.


fran-worley:
@apotonick can you have a look at this? The failing test is due to the key format of errors on a collection. I just need to have a look at how AM-V output them and then we should be good to go.

fran-worley:
Ok so I've sorted out how the errors should map to objects and return consistently with the AM-V api.

My problem now is that for collections I need the fragment index in order to map the dry error to the correct collection item. @apotonick  is there a way that we can store the fragment index in a collection form object for me to hook into?

Currently you have to pass it into the params hash for it to work which isn't ideal.

apotonick:
@fran-worley Re. the index, you could "hack" it with

```ruby
form.parent.songs.index(form)
```

BUT... we can make that simpler and set the index on the form itself or something like that.

rosskevin:
Just updated, getting the following error:

```
expected Trailblazer::Operation::InvalidContract, got #<NoMethodError: undefined method `fragment_index' for #<#<Class:0x007fab3a6c3d28>:0x007fab3c9b8298>> with backtrace:
  # gems/reform-c4fc02510914/lib/reform/form/dry.rb:83:in `block (2 levels) in add_nested_error_message'
  # gems/disposable-0.4.0/lib/disposable/twin/property_processor.rb:24:in `block in collection!'
  # gems/disposable-0.4.0/lib/disposable/twin/property_processor.rb:24:in `collect'
  # gems/disposable-0.4.0/lib/disposable/twin/property_processor.rb:24:in `collection!'
  # gems/disposable-0.4.0/lib/disposable/twin/property_processor.rb:15:in `call'
  # gems/reform-c4fc02510914/lib/reform/form/dry.rb:81:in `block in add_nested_error_message'
```

Need more info?

fran-worley:
@rosskevin because I'd forgotten you were actually using this branch and that change is a bit of a hack right now. I'll sort this week and hopefully merge.

fran-worley:
@rosskevin I've removed the hack you will need to make sure that you're adding:

    require "disposable/twin/parent" 
    feature Disposable::Twin::Parent 

to your forms with collections or you'll get errors about an unknown parent method.

fran-worley:
As I've changed the API for groups (again :)) we will need to merge this PR https://github.com/trailblazer/reform-rails/pull/23 in reform-rails aswell.

fran-worley:
So cool feature for reform/dry-v added. You can now inject your own dependencies into a schema without having to make them available via a method on your form. You can do so by passing in a `:with` option to your validation group as follows:

```ruby
class SomeForm < Reform::Form
  validation :default, with: { user: current_user } do
    configure do
      option :user # as per dry-v docs, you must declare injected options in the configure block
      
      def users_name
        user.name # I can now access my user Yey!
      end
    end
    required(:username).filled(eql?: users_name)
  end
end
```

rosskevin:
Hello @fran-worley - I've got a spec where I'm expecting `InvalidContract`, but I'm getting an error on this branch:

```ruby
expected Trailblazer::Operation::InvalidContract, got #<NoMethodError: undefined method `parent' for #<#<Class:0x007f8ebd638938>:0x007f8ebfd4fa58>
Did you mean?  present?> with backtrace:
  # /Users/kross/.rvm/gems/ruby-2.3.1@af/bundler/gems/reform-89505ae12a11/lib/reform/form/dry.rb:86:in `block (2 levels) in add_nested_error_message'
  # /Users/kross/.rvm/gems/ruby-2.3.1@af/gems/disposable-0.4.0/lib/disposable/twin/property_processor.rb:24:in `block in collection!'
  # /Users/kross/.rvm/gems/ruby-2.3.1@af/gems/disposable-0.4.0/lib/disposable/twin/property_processor.rb:24:in `collect'
  # /Users/kross/.rvm/gems/ruby-2.3.1@af/gems/disposable-0.4.0/lib/disposable/twin/property_processor.rb:24:in `collection!'
  # /Users/kross/.rvm/gems/ruby-2.3.1@af/gems/disposable-0.4.0/lib/disposable/twin/property_processor.rb:15:in `call'
  # /Users/kross/.rvm/gems/ruby-2.3.1@af/bundler/gems/reform-89505ae12a11/lib/reform/form/dry.rb:84:in `block in add_nested_error_message'
  # /Users/kross/.rvm/gems/ruby-2.3.1@af/gems/disposable-0.4.0/lib/disposable/twin/definitions.rb:22:in `block in each'
  # /Users/kross/.rvm/gems/ruby-2.3.1@af/gems/declarative-0.0.8/lib/declarative/definitions.rb:30:in `each'
  # /Users/kross/.rvm/gems/ruby-2.3.1@af/gems/declarative-0.0.8/lib/declarative/definitions.rb:30:in `each'
  # /Users/kross/.rvm/gems/ruby-2.3.1@af/gems/disposable-0.4.0/lib/disposable/twin/definitions.rb:16:in `each'
  # /Users/kross/.rvm/gems/ruby-2.3.1@af/bundler/gems/reform-89505ae12a11/lib/reform/form/dry.rb:81:in `add_nested_error_message'
  # /Users/kross/.rvm/gems/ruby-2.3.1@af/bundler/gems/reform-89505ae12a11/lib/reform/form/dry.rb:70:in `block in process_errors'
  # /Users/kross/.rvm/gems/ruby-2.3.1@af/bundler/gems/reform-89505ae12a11/lib/reform/form/dry.rb:68:in `each'
  # /Users/kross/.rvm/gems/ruby-2.3.1@af/bundler/gems/reform-89505ae12a11/lib/reform/form/dry.rb:68:in `process_errors'
  # /Users/kross/.rvm/gems/ruby-2.3.1@af/bundler/gems/reform-89505ae12a11/lib/reform/form/dry.rb:52:in `call'
  # /Users/kross/.rvm/gems/ruby-2.3.1@af/bundler/gems/reform-89505ae12a11/lib/reform/validation/groups.rb:57:in `block in call'
  # /Users/kross/.rvm/gems/ruby-2.3.1@af/bundler/gems/reform-89505ae12a11/lib/reform/validation/groups.rb:51:in `each'
  # /Users/kross/.rvm/gems/ruby-2.3.1@af/bundler/gems/reform-89505ae12a11/lib/reform/validation/groups.rb:51:in `call'
  # /Users/kross/.rvm/gems/ruby-2.3.1@af/bundler/gems/reform-89505ae12a11/lib/reform/validation.rb:39:in `valid?'
  # /Users/kross/.rvm/gems/ruby-2.3.1@af/bundler/gems/reform-89505ae12a11/lib/reform/contract/validate.rb:18:in `validate!'
  # /Users/kross/.rvm/gems/ruby-2.3.1@af/bundler/gems/reform-89505ae12a11/lib/reform/contract/validate.rb:10:in `validate'
  # /Users/kross/.rvm/gems/ruby-2.3.1@af/bundler/gems/reform-89505ae12a11/lib/reform/form/validate.rb:24:in `validate'
  # /Users/kross/.rvm/gems/ruby-2.3.1@af/gems/trailblazer-1.1.0/lib/trailblazer/operation.rb:144:in `validate_contract'
  # /Users/kross/.rvm/gems/ruby-2.3.1@af/gems/trailblazer-1.1.0/lib/trailblazer/operation.rb:134:in `validate'
  # /Users/kross/.rvm/gems/ruby-2.3.1@af/gems/trailblazer-1.1.0/lib/trailblazer/operation/model.rb:36:in `validate'
  # ./app/concepts/af/commerce/base_itemized_document/operation.rb:24:in `process'
``` 

Other than that, this branch is working well for the rest of my codebase.

fran-worley:
Hi Kevin,

As per: https://github.com/apotonick/reform/pull/388#issuecomment-244919594 you'll need to include/ feature Parent when using collection validation via dry-v.

rosskevin:
Thank you @fran-worley, I'm sorry I missed that.

I included it and now receive this:

```ruby
/Users/kross/.rvm/gems/ruby-2.3.1@af/gems/disposable-0.4.0/lib/disposable/twin/parent.rb:3:in `included': private method `property' called for #<Class:0x007fb3ffe194b0> (NoMethodError)
Did you mean?  properties
	from /Users/kross/.rvm/gems/ruby-2.3.1@af/gems/declarative-0.0.8/lib/declarative/schema.rb:72:in `include'
	from /Users/kross/.rvm/gems/ruby-2.3.1@af/gems/declarative-0.0.8/lib/declarative/schema.rb:72:in `block in feature'
	from /Users/kross/.rvm/gems/ruby-2.3.1@af/gems/declarative-0.0.8/lib/declarative/schema.rb:71:in `each'
	from /Users/kross/.rvm/gems/ruby-2.3.1@af/gems/declarative-0.0.8/lib/declarative/schema.rb:71:in `feature'
	from /Users/kross/projects/af/app/concepts/af/commerce/base_itemized_document/operation.rb:16:in `block in <class:Create>'
```

Here's a contract snippet:

```ruby
        contract do
          feature Disposable::Twin::Parent
          collection :line_items, populate_if_empty: LineItem do
            property :product_id
            property :quantity
            property :period_start_at
            property :period_end_at
          end
        end
```

fran-worley:
@rosskevin how weird this test passes fine:
https://github.com/apotonick/reform/blob/improve-dry-integration/test/validation/dry_validation_test.rb#L247-L356

Going to have to call in the bid Disposable guns here... Any ideas @apotonick  ??

rosskevin:
Just to be sure I'm not leaving out anything useful, it is breaking on the line that calls `feature Disposable::Twin::Parent`, which wouldn't make sense if the test is running.  I'll check out this branch and look at dependency versions.

rosskevin:
Tests run locally and nothing sticks out on the surface dependency-wise.  I updated the error above with the full stack.

fran-worley:
Latest commits remove that default `:form` dependency injection into dry schemas. This also removes the `Reform::Form::Schema` class as it was no longer needed. 

To manually inject your form instance into your schema, you just need to add `with: { form: true }` option to your validation call (obviously not forgetting to add `option :form` to your dry-v configure block).

test/validation/dry_validation_test.rb
```diff
@@ -363,6 +363,21 @@ def good_musical_taste?(value)
...
+    it "renders full messages correcty" do
+      result = form.validate(
+        "title"  => "",
+        "songs"  => [ {"title" => ""}, {"title" => ""} ],
+        "band"   => {"size" => "", "label" => {"name" => ""}},
+        "producers" => [{"name" => ''}, {"name" => 'something lovely'}]
+      )
+
+      result.must_equal false
+      form.band.errors.full_messages.inspect.must_equal %({:name=>["Name must be filled"], :\"label.name\"=>[\"Label Name must be filled\"]})
```
fran-worley: Thanks for pointing this out. I'll confess that I've never actually used ActiveRecords full messages and couldn't remember exactly how they came out. 

Does this look better? https://github.com/trailblazer/reform/tree/fix-arstyle-full-messages

JoeWoodward: Hey Fran, thanks for hopping on that so quickly. I don't see any changes in the branch you linked to? Did you forget to push?

I've just looked at the implementation and realised that I think it's actually not your fault at all. It's due to the way dry-validation formats the full_messages

JoeWoodward: https://github.com/trailblazer/reform/blob/330f6b6d03eba129e7792ab3bf5e90433744d4e7/lib/reform/form/dry.rb#L13

http://dry-rb.org/gems/dry-validation/error-messages/

contentfree:
So do usages of `optional()` in a validation scheme get treated as expected from dry-v docs? (I saw the issue got rolled into this one, but it looks like it was unresolved.)

apotonick:
@contentfree Any example how that would be used and expected outcome?

fran-worley:
@apotonick this was the purpose of the context option I added.

The issue with Reform is that if the property is declared on the form the key gets included regardless of whether it was passed in as a param.

If the key is then passed to dry then the optional key concept in dry is unusable with Reform.

apotonick:
The solution here is passing only submitted attributes to dry?

fran-worley:
But then what about default properties? Making it configurable might be better?

contentfree:
I think the least-surprising experience for a dev would be for the optional/required abilities stated in the dry-v docs to work as described *and* for properties with defaults to work as expected.

siassaj:
> I think the least-surprising experience for a dev would be for the optional/required abilities stated in the dry-v docs to work as described and for properties with defaults to work as expected.

Thing is, we're either validating input params or the result of the form being populated. Can't really have it both ways with 1 set of validations. To be honest I'm close to declaring that reform::forms can have their own validation block which makes sure the data makes sense before sync, and operations can have their own validation block which makes sure the passed params make sense.

@apotonick whatcha think?

siassaj:
As a follow up; this terrible stuff:

'''ruby
property :location_id, populator: -> (opts) { self.location_id = nil }
required(:location_id).filled(:str?)

Order::Checkout.({location_id: "hi"})[:form].errors
=> #<Reform::Contract::Errors:0x00556d20c1db48
 @errors=
  {:location_id=>["must be filled"],
'''

apotonick:
1. The idea in Reform has always been to validate the "final shape of the form" (good point @siassaj ), not the user input. This now is a problem because of the way dry-v works. Solution is to only pass changed/defaulted (good point @contentfree  ) params to the dry-v engine.

2. With regards to orchestrating the flow of a validation, I have a massive addition to Reform planned, where the whole validation/deserialization/coercion/etc. will be an operation itself that can be reordered and the like. This will dramatically simplify how Reform works. Since TRB 2.1 is around the corner, I am expecting to find some time crafting a draft (and a draught) next week.

siassaj:
> The idea in Reform has always been to validate the "final shape of the form"
 
So I'm a bit torn; ultimately we expect to sync the form data into another domain, whereupon that domain should take care of what's valid or not (simple example, activerecord validations for what ultimately should/shouldn't go into db, or even better postgres triggers and rules etc.).

Then comes the idea of what flows into the form, that is then combined into the final shape of the form via populators, defaults, etc. In the simplest case this may be a couple populator lambdas, some defaults and the final params passed into the form.

In my thinking these sources of data ought to either be valid already, or validated by the form. Here's where I'm torn;
  These form objects back the actual form the user fills in, i want to validate the users input. It's the only source of data that could be invalid in almost every case I see. Incorrect defaults/populators and other sources of data ... they feel like something to be tested for.
  And it's not a trivial situation; on one hand we have a form with raw user input, that raw stuff needs validating in my mind. On the other hand we have domain objects that need to be valid before we use/store them. The form being in a valid shape before we sync also makes sense. But now we have a mismatch against. validating user input

apotonick:
That's exactly the "dilemma" we're in right now! Dry-v is really only meant for asserting input. Reform's statement is that we need higher-level business validations, too (e.g. if cart.items.count > 3 total amount needs to be > $100). To solve this, those two might have to get separated, which is why I want to introduce the validation "flow".

siassaj:
Yep makes sense, We need validation flow.

I'm gonna wait for you to write that :P in the meantime i'll just hack reform::form to have a 2 step validation process :D

contentfree:
You can always make two contracts: One for the input and one for the model. Though I do agree with something @siassaj implied: ActiveRecord objects should be where its validation should reside. Ultimately, it's the thing responsible for syncing your data store (in your app… validate at the DB, too)

Issue #367, opened 2016-06-08
Setting skip_if: :all_blank on collections prevents the nested models from being persisted


cattekin:
when used on a collection for a `has_many`, `skip_if: :all_blank` is preventing the nested collection from being persisted, even with valid attributes.

for a rails app demoing this: [example rails app](https://github.com/cattekin/reform_collection)

sample collection code which produces this error:

``` ruby
class PostsForm < Reform::Form
  property :body, validates: { presence: true }

  collection :comments, form: CommentsForm,
    skip_if: :all_blank,
    populator: -> (collection:, index:, **) do
      if item = collection[index]
        item
      else
        collection.append(Comment.new)
      end
    end
end
```

apotonick:
Can you give an example what the collection looks before and after the `validate` call, and also paste the incoming params?

cattekin:
Hey @apotonick, I dug into this on Tuesday, main cause looks to be trying to compare strings / symbols in the AllBlank class.

the keys for `options[:input]` are passed in as symbols,
but the names in `options[:binding][:nested].definitions.collect { |dfn| dfn[:name] }` are strings

If you still want the collection before after or params, let me know.
Also happy to do a small PR with a fix.

added the code with comments:
``` ruby
# lib/reform/form/validate.rb
module Skip
  class AllBlank
    include Uber::Callable

    def call(form, options)
      # keys here are passed in as symbols
      params = options[:input]
      # definition names are passed in here as strings
      properties = options[:binding][:nested].definitions.collect { |dfn| dfn[:name] }

      properties.each { |name| params[name].present? and return false }
      true # skip
    end
  end
end
```

fran-worley:
@cattekin a PR with a test case and a fix would be lovely if you're still up for it.

As of Reform 2.2 we moved all rails code to reform-rails so please don't use `.present?` or  other ActiveSupport methods in your PR.

cattekin:
@fran-worley pushed a PR for this.
Let me know if it needs any work

emaglio:
Fixed by https://github.com/trailblazer/reform/pull/392

PR #392, opened 2016-08-13
Handle strings or symbols as key for AllBlank

cattekin:
Ensure AllBlank returns correctly if forms have content, regardless of
type of key used.

Fixes https://github.com/apotonick/reform/issues/367#issuecomment-238605908

The explicit type conversion could be removed if the keys being passed in are guaranteed to be strings, but I'm not familiar enough with Dry schemas to attempt that currently.

### DIFF
--- a/lib/reform/form/validate.rb
+++ b/lib/reform/form/validate.rb
@@ -5,12 +5,19 @@ module Reform::Form::Validate
       include Uber::Callable
 
       def call(form, options)
-        params = options[:input]
         # TODO: Schema should provide property names as plain list.
-        properties = options[:binding][:nested].definitions.collect { |dfn| dfn[:name] }
+        # ensure param keys are strings.
+        params = options[:input].each_with_object({}) { |(k, v), hash|
+          hash[k.to_s] = v
+        }
 
-        properties.each { |name| (!params[name].nil? && params[name] != "") and return false }
-        true # skip
+        # return false if any property inputs are populated.
+        options[:binding][:nested].definitions.each do |definition|
+          value = params[definition.name.to_s]
+          return false if (!value.nil? && value != '')
+        end
+
+        true # skip this property
       end
     end
   end
--- a/test/skip_if_test.rb
+++ b/test/skip_if_test.rb
@@ -71,4 +71,11 @@ class SkipIfAllBlankTest < BaseTest
     form.songs.size.must_equal 1
     form.songs[0].title.must_equal "Apathy"
   end
+
+  it do
+    form = AlbumForm.new(OpenStruct.new(songs: []))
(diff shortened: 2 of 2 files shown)


fran-worley:
Thanks for this. Can you expand your test to cover nested properties and collections? 

I'll have a think as to whether there is a better fix as recursively converting another hash to strings is a pain and there must be a better way! 

Are you able to isolate when this happens? I'm guessing it's something to do with whether you pass the params in as strings of symbols in the first place.

apotonick:
@cattekin Thanks!! As @fran-worley pointed out, a stronger (nested) test would be cool. Can you push the test in a separate commit? We will then merge the test, but I will change the implementation as Representable 3.0.1 will support the symbol/string conversion. This will reduce code in Reform.

cattekin:
@apotonick will add some more tests and split out the specs to a seperate commit.

fran-worley:
@cattekin are you able to add the extra tests, or should we take over this one? I've bumped into it myself a couple of times and I'd quite like to get it merged you see!

cattekin:
@fran-worley Sorry, thought this PR had been closed.
I've split the spec changes out into a seperate commit.

Any examples of the type of specs you'd like in the codebase?

fran-worley:
@cattekin 🌟 !!

I just wanted you to expand the form used in your test case to cover deeply nested objects, something like this (obvs you'll need to add skip_ifs and populators etc.)

```ruby
class AlbumForm < TestForm
  property :artist do
    property :name
    # test that a collection inside a nested property works
    collection :members, skip_if: :all_blank do
      property :name
    end
  end

  collection :songs, skip_if: :all_blank, populate_if_empty: BaseTest::Song do
    property :title
    property :length
    # test that a collection within a collection works
    collection :featuring_artists, skip_if: :all_blank do
      property :name
    end
  end
end
```

emaglio:
@fran-worley @apotonick should we consider these changes for the next release - if yes, let’s add it in the 2.3v milestone

SirRawlins:
Caught by this myself today, banging my head against the desk trying to understand why `skip_if: :all_blank` wasn't working, only to find it was due to symbol-based keys. 🙈 

Any plans for this to make it's way in?

emaglio:
@SirRawlins going to investigate if it will be something easy will be added in the release of 2.3.0 otherwise will be in the 3.0 (no estimated time on this release).
I'll update the issue in base on the above :+1:

emaglio:
@SirRawlins this is going to be in the next release of reform `2.3.0.rc2` for now can you please test `v2.3.0.rc2` branch?

SirRawlins:
@emaglio - thanks so much! 👍 I'll try and give that a proper test but initial glance it all looks good.

PR #393, opened 2016-08-23
Add alias to writeable property option.

sot528:
I did mistake of spelling as `writable` to `writeable` property option. 😂 
Same as [Writeable vs writable](https://github.com/apotonick/representable/issues/188)

I think it's a little bit dengerous, so I added alias.


lib/reform/form.rb
```diff
@@ -21,6 +21,10 @@ def property(name, options={}, &block)
           options[:deserializer][:writeable] = options.delete(:parse)
         end
 
+        if options.key?(:writable)
+          options[:writeable] = options.delete(:writable)
```
fran-worley: Couldn't we just do: `options[:writeable] ||= options.delete(:writable)`

test/writeable_test.rb
```diff
@@ -26,4 +26,33 @@ class LocationForm < Reform::Form
 
     hash.must_equal("country"=> "Germany")
   end
-end
\ No newline at end of file
+end
+
+# writable option is alias of wiriteable option.
```
fran-worley: If I'm being super picky, there's a type in your comment... 😜 

wiriteable => writeable

fran-worley:
@sot528 thanks for this, a very good spot. It's an odd one as the Oxford English dictionary doesn't recognise writeable as a word but my laptop spell check doesn't recognise writable as a word! Can't win so lets just support both.

fran-worley:
Thanks @sot528 :shipit:

sot528:
@fran-worley 
Thank you for merging quickly. 😉

test/writeable_test.rb
sot528: 😂

RKushnir:
@sot528 you may also need an alias for the word "dengerous" ;)

sot528:
@RKushnir That's true 😂

audionerd:
On the latest released reform gem 2.2.3, `writable` is still silently ignored.

On `master`, `writable` is supported as well as `writeable`.

sot528:
@audionerd 
Thank you, I checked your indication.
As a result, it's included `v2.3.0`, not until `v2.2.4`. :) 
- https://github.com/trailblazer/reform/compare/v2.2.4...v2.3.0.rc1

PR #399, opened 2016-09-15
Add default params to validate method

erichin:
In some case,  if we have set all the property, there is no need to set the property again from params before property validation. So the validate method need a default parameter for the direct use without params, such like
```
form.validate
```


fran-worley:
@erichin any reason you keeping opening and closing this? Do you want us to consider it?

PR #411, opened 2016-12-03
Clean up and restore some inherit_test code.

lukeasrodgers:
The `subject.validate` tests can be restored with some fairly minor syntax changes to be compatible with dry-validation.

There may be reasons to keep some of these tests commented-out that I'm not aware of.

The motivation for this clean-up is that I'm encountering some odd behaviour (possibly a bug, possibly a misunderstanding on my part) around inherited populators, and having a more thorough set of tests here is a prerequisite for me writing some tests to verify whether my issue is a bug or not.

cheers

### DIFF
--- a/test/inherit_test.rb
+++ b/test/inherit_test.rb
@@ -4,72 +4,66 @@ require 'representable/json'
 class InheritTest < BaseTest
   Populator = Reform::Form::Populator
 
+  class SkipParse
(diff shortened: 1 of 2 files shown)


fran-worley:
Thanks for this. 👍

apotonick:
Cool, looking way better now! :heart_eyes:

Issue #345, opened 2016-03-10
Error Message Api


dlupu:
Hello

According to the [documentation](http://trailblazer.to/gems/reform/#api)  `#errors returns validation messages in a classic ActiveModel style``

This does not seem to be completely true when using Dry Validations. 

For instance `errors.full_messages` will generate `NoMethodError: undefined method 'full_messages' for #<Reform::Contract::Errors:0x007fecf8c3f798 @errors={}>`

Same happens for `errors.full_message_for(:attribute)`. 

SInce reform is planning to drop support for Active Model Validations, I think that providing a stable errors API would facilitate the migration to dry validations. WDYT ?  

BR
Dorian

jesseiqmi:
Agreed and more so to the point that I just finished running the gauntlet on getting Reform implemented and now just fine tuning my first form that uses a collection and now dealing with errors being displayed back to the user.  I'm using `reform` and `reform-rails` 2.1 with Rails 4.2.x. 

When I drop in a custom validator for the collection and using `errors.add(:something, 'at least one is required)` it gets added first, then as the rest of the reform validations kick in they get added afterwards.

Example:
```ruby
property :title
property :date

validates :title, presence: true
validates :date, presence: true
validate :validate_at_least_one_foo

collection :foos do
  property :thing1
  property :thing2
end

def validate_at_least_one_foo
  errors.add(:foos, 'at least one is required') if foos.size == 0
end
```

Errors as displayed to user if all are triggered:

* Foos - at least one is required
* Thing 1 - is required
* Thing 2 - is required

Any thoughts on how we can control the display order of errors?  Ideally, as this deals with a collection of 'foos' at the bottom of the UI page, its order in the error messages as displayed to the user should appear last, not first so that they can address issues from the top of the page first and work down.

FYI I jumped on to this ticket as the behavior is different from Rails handling of errors - I can created a new ticket if better?

Cheers and thanks!

dlupu:
@jesseiqmi I think you should have created a new ticket, we're not speaking about the same problem.

jesseiqmi:
Thanks @dlupu after rereading you're specifically talking about dry-validations.

I've moved this out to its own ticket https://github.com/apotonick/reform/issues/351

apotonick:
The `Errors` API with dry-v is still subject to change, it'll be compatible with the one we have now!

fran-worley:
[...]

PR #416, opened 2016-12-11
Add `Errors#count` and `#size` methods.

lukeasrodgers:
This improves compatibility with ActiveModel-style errors (which seems to be a goal of the gem, according to the docs and #345), and also makes integration with some gems other gems easier.

For example, some of the views for the devise authentication gem expect that an `errors` object will respond to `count`.

Obviously AM compatibility could be pushed further, but these changes are fairly innocuous and straightforwardly helpful.

### DIFF
--- a/lib/reform/contract/errors.rb
+++ b/lib/reform/contract/errors.rb
@@ -49,4 +49,13 @@ class Reform::Contract::Errors
   def [](name)
     @errors[name] || []
   end
+
+  def size
+    @errors.values.flatten.size
+  end
+  alias :count :size
+
+  def empty?
+    size.zero?
+  end
 end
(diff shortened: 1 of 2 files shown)


apotonick:
Thanks Luke. @fran-worley will this be compatible with the Dry errors structure we're working on? Is there such a concept as "size" for their errors?

@lukeasrodgers The goal of this gem is to get rid of as much Active* as possible, so any method that makes Reform work better with Rails should go to reform-rails.

lukeasrodgers:
@apotonick got it, thanks. Feel free to close if this is not compatible with planned changes to the API. 

That said, I think that regardless of AM compatibility, a `count` method on an errors object would be helpful -- it looks like dry-validation gets this via including `Enumerable`, however this does not seem to be directly compatible with my change here, since dry-rb's errors object is (afaict) just a hash, so `count` is going to count keys rather than values.

IMHO I think the count on `values` is more intuitive (e.g. if there are two fields, each with two errors, the error count should be 4, not 2).

fran-worley:
Can we make one amendment to remove the alias for .blank? And then merge. I'll revisit which gem this ends up in later :)

lukeasrodgers:
@fran-worley done, squashed into one commit.

PR #418, opened 2016-12-16
Fix arstyle full messages

JoeWoodward:
@fran-worley this looks good now, all is working correctly


JoeWoodward:
One thing I would like to mention. When returning a full_message for a property that is aliased, the property name is returned not the alias, is that correct?

If so, is there any way to specify I would like to use the alias name?

PR #420, opened 2016-12-21
Update stale comment about use of representable gem.

lukeasrodgers:
Unless you've manually opted into it, the call to `super` here will go through the disposable and declarative gems, but not representable.

This is obviously a very minor change, but caused a bit of confusion for me when attempting to understand the source code. 

Of course, my understanding may be wrong, feel free to close the PR if it is.

### DIFF
--- a/lib/reform/form.rb
+++ b/lib/reform/form.rb
@@ -31,7 +31,7 @@ module Reform
           options[:writeable] ||= options.delete(:writable)
         end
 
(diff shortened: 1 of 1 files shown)


apotonick:
This is correct, but please spell `declarative` correctly. :stuck_out_tongue_closed_eyes:

lukeasrodgers:
d'oh :)

PR #435, opened 2017-05-25
added dry-v logic operator tests to demonstrate quirks in DSL

jodeci:
Came across two issues:

- (named) logic operators half supported

Not sure if this is intended behaviour, because `and`, `or` can be used with no problem. Anyway, I meant to submit a failing test, but using `xor`, `then` results into syntax errors so you'll need to tweak the syntax manually (L:242, L:268).
 
- xor not exclusive

Hopefully the failing test speaks for itself (L:258).

### DIFF
--- a/test/validation/dry_validation_test.rb
+++ b/test/validation/dry_validation_test.rb
@@ -237,6 +237,9 @@ class DryValidationLogicOperatorTest < MiniTest::Spec
       property :bar
 
       validation do
+        required(:foo).filled
+        required(:bar).filled
+
         rule(foobar: [:foo, :bar]) do |foo, bar|
           # syntax error when using "xor" alias
           foo.eql?("foobar") ^ bar.eql?("foobar")
@@ -254,7 +257,6 @@ class DryValidationLogicOperatorTest < MiniTest::Spec
 
     # invalid
     it do
-      # should pass but fails
       form.validate({ foo: "foobar", bar: "foobar"}).must_equal false
     end
   end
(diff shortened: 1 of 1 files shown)


apotonick:
@jodeci Could you find out what Dry does to "allow" those DSL quirks? Because, Reform doesn't do anything special but simply `eval`uating the Dry block in its own class. I have zero idea what to change here?

Cheers from Pingtung :sunny:

fran-worley:
I think you have to declare a base rule (e.g. required(:key).maybe) for any keys you want to base a high level rule off.
This isn't a Reform issue but a confusing quirk with Dry-Validation.

jodeci:
@fran-worley ah good catch, the xor test is now passing!

@apotonick I'll try to dig deeper in dry-v, but so far my understanding is that the named operations are nicely defined with their `&|^>` aliases here and should "just work":

https://github.com/dry-rb/dry-validation/blob/master/lib/dry/validation/schema/rule.rb#L152

And let me buy you a beer if you're stopping by Taipei! 🍺

jodeci:
Yeah... so apparently it's a dry-v issue.

https://github.com/dry-rb/dry-validation/blob/master/spec/integration/schema/xor_spec.rb#L17
https://github.com/dry-rb/dry-validation/blob/master/spec/integration/injecting_rules_spec.rb#L6

Changing `>`, `^` to `then`,`xor` give syntax errors too. Gonna open an issue there!

PR #448, opened 2017-12-19
add a test for errors generated by dry-v `each` macros

Tab10id:
Test failed on 2.3 but it work with 2.2


fran-worley:
@Tab10id you'll be pleased to know that I've just bumped into this too so it's likely to get resolved fairly quickly!!

fran-worley:
Fix added here - https://github.com/trailblazer/reform/commit/a7c1bee4d8679fc070c87ff80b47d9a662b08724

Note that this may need to be revisited for simple_form/ rails form builder support as I'm not sure the format is correct.

PR #458, opened 2018-05-25
Update licence

seuros:


emaglio:
@seuros this can be closed, right?

Issue #410, opened 2016-11-29
Expose raw validation result


dnd:
There is currently no way to access the raw validation result that is returned from dry-validations, and it contains additional information that can be used to create more complex/detailed error returns. By exposing this value users can still continue to use `Form#errors` for the simple error messages, or they can access the raw result (`Form#validation_result` maybe?) to construct a more elaborate response.

fran-worley:
It would be useful to know what extra data it is that your looking for.

If you're trying to acces hints vs errors or full messages vs standard then oh won't need the raw result. Off the top of my head I can't think what else youd need as your input is essentially form.to_nested_hash...

dnd:
Hints, and errors yes. I'm trying to also get changes made to the MessageSet returned from dry-validation to include additional information in it as well. Ideally I'm looking for a return along the lines of this: https://discuss.dry-rb.org/t/different-data-types-for-errors-on-array-of-hashes/152/3?u=dnd

Yes, I want the errors(messages), and the hints, but I'm also looking for the predicate value to provide more  machine friendly errors messages. So the data structure that Reform uses would have to change completely to support more complex values for hints and messages to have the predicate, and potentially tokens in the future, as right now it's just field/string pairs.

apotonick:
I refactored the errors code last night and we can now transport those raw errors. It will be an array, though, one for every validation group. Is that still helpful, @dnd ?

apotonick:
@dnd See here: https://github.com/trailblazer/reform/blob/nick-simple/lib/reform/contract/errors.rb#L2

apotonick:
Once the dry errors format becomes stable, we could think about exposing it directly via Reform's `Errors`.

apotonick:
You can grab the original Dry errors object with the `Form#to_result` method. It's undocumented and considered private, but will do the trick!

dnd:
@apotonick sorry to dig this up to the dead, but I finally get to work with Trailblazer again. :) `#to_result` almost gets me there. I still have to do `form.to_result.instance_variable_get(:@results)` or `:@failure` to actually get the raw dry-validation data. 

Would you accept a PR to the `2.3.0.rc2` branch to make readers for `@failure` and `@results` on `Reform::Contract::Result`? Maybe even label the readers `raw_failure`, or `raw_results` so that it is explicit that they're not part Reform?

All the data I need is in there to construct the API errors I need. I just don't want to have to reach in an grab instance variables for obvious reasons.

[...]

PR #463, opened 2018-08-17
Expose raw validation results

dnd:
As per @apotonick's request in #410 here is a PR to start the conversation about changes to Reform.

A couple things I noticed as I went through this.
- `#failure?` did not actually return a boolean value. It just returned whatever was in `@failure`. Since a new `#failure` method was being added, I updated `#failure?` to be more consistent with the expectation of a `?` method.
-  Is the `result.rb` line 9 is correct behavior? Should it instead be a `map`? As it is now, it's just going to find the first failure of the combined lists. Couldn't there be multiple?


dnd:
@apotonick I don't think these failures are the fault of anything I did, are they?

emaglio:
@fran-worley @apotonick should we consider these changes for the next release - if yes, let’s add it in the 2.3v milestone

PR #481, opened 2019-08-08
Warn user when validate is called more than ones

emaglio:
This will notify the User if a double validate happens (https://github.com/trailblazer/reform/issues/447) - discarded the solution to raise errors since it can break a lot of people code even though it's a wrong solution anyway.

Adding a note in the 3.0 project where we make sure to handle this in a better way


emaglio:
This is going to be fix in 3.0

PR #494, opened 2019-09-13
Allow to add an error after validation

emaglio:
The idea is to add in the `@results` instance variable an object which simulates a Dry schema object so any other part of the code should just work as before.
I have also added the possibility to add an error before actually running a validation (not sure if that makes sense but I don't see why it wouldn't).

### DIFF
--- a/lib/reform/contract/custom_error.rb
+++ b/lib/reform/contract/custom_error.rb
@@ -1,0 +1,35 @@
+module Reform
+  class Contract < Disposable::Twin
+    # a "fake" Dry schema object to add into the @results array
+    # super ugly hack required for 2.3.x version since we are creating
+    # a new Reform::Errors instance every time we call form.errors
+    class CustomError
+      def initialize(key, error_text, results)
+        @key        = key
+        @error_text = error_text
+        @errors     = {key => [error_text]}
+        @messages   = @errors
+        @hint       = {}
+        @results    = results
+
+        merge!
+      end
+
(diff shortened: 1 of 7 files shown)


lib/reform/contract/custom_error.rb
```diff
@@ -0,0 +1,35 @@
...
+      end
+
+      def failure?
+        true
+      end
+
+      def merge!
+        @results.map(&:errors)
+                .detect { |hash| hash.key?(@key) }
+                .tap { |hash| hash.nil? ? @results << self : hash[@key] |= [@error_text] }
```
emaglio: this is basically get all the errors in `@results`, check if any of the error has an has with `@key` if it has then merge the error texts using `|=` otherwise simply add `self` into the `@results` variable.
Better way to do so?

lib/reform/contract/validate.rb
```diff
@@ -33,7 +33,7 @@ def validate!(name, pointers = [])
       nested_errors = validate_nested!(pointers_for_nested)
 
       # Result: unified interface #success?, #messages, etc.
-      @result = Result.new(local_errors_by_group + pointers, nested_errors)
+      @result = Result.new(@result.to_result + local_errors_by_group + pointers, nested_errors)
```
emaglio: this is the part where we add `CustomError` before `validate` is called - by default that will return `[]`

apotonick:
:+1:

PR #495, opened 2019-09-14
Reintroduce valid?

emaglio:
Calls `validate({})`.

The aim of this method is to check if that specific Form instance is valid if model changes a new form instance MUST be created.
We have issues around about calling multiple times validate to the same instance and we know that has problems - this will be fixed in the 3.x


apotonick:
As bespoke, in Reform 3 the API won't allow validating twice since there won't be an instance to validate but something more like `result = Form.validate(presented_form, params)`.

emaglio:
@apotonick this will give us compatibility with 2.2.4 and then from 3.0 we will require bigger changes for an upgrade

PR #496, opened 2019-09-16
Add compatibility with dry-v 1.x

emaglio:
The main changes are due to a quite different API for newer versions of dry-v.
The real changes are 2:
- the way `validation_group` is initialized and called
  - validations `options` are ONLY dynamically added in the schema on `CALL` since the DSL calls `instance_eval` on `new` which will raise an error if any of the methods are not implemented 
- the result of a dry-v call is not just an hash anymore so we need to use `to_h` to recreate the same structure as before

Apart from those changes above the rest is just to be able to support 2 major versions of dry in the same gem.

Following the commit history will make the review a bit easier

**From Reform 3.x we will drop dry-v < 1.x**


lib/reform/form/dry.rb
```diff
@@ -1,97 +1,19 @@
...
-        Group
-      end
-    end
-
-    def self.included(includer)
-      includer.extend(ClassMethods)
+    if Gem::Version.new(Dry::Validation::VERSION) > Gem::Version.new("0.13.3")
+      require "reform/form/dry/new_api"
+      validations = Reform::Form::Dry::NewApi::Validations
+    else
```
apotonick: Shouldn't we just require Dry >= 1.0 in 2.3?

emaglio: I'm quite sure there are a lot of people with 2.3.0.rcx and dry-v 0.13 so I think we should support both and only 1.x for reform 3.x
What do you think?

apotonick: Since we're throwing away most of Reform 2.x soon, yes, you're right! :pray:

emaglio: Still in draft here so we can discuss a better solution maybe to require files for a specific file version (same for the tests) but yeah for now I want to have it working and then I can improve.
That is why the PR is still in DRAFT :+1:

lib/reform/form/dry/input_hash.rb
```diff
@@ -0,0 +1,37 @@
...
+    # we can't use to_nested_hash as it get's messed up by composition.
+    def input_hash(form)
+      hash = form.class.nested_hash_representer.new(form).to_hash
+      symbolize_hash(hash)
+    end
+
+    # dry-v needs symbolized keys
+    # TODO: Don't do this here... Representers??
+    def symbolize_hash(old_hash)
+      old_hash.each_with_object({}) do |(k, v), new_hash|
```
apotonick: I think I saw some Dry-specific code somewhere in `dry-equalizer` or something to do that, maybe @solnic can help?

solnic: @apotonick dry-validation symbolizes hashes for `Params` and `JSON` schemas

emaglio: I believe we need to keep this for old versions of dry-v?

test/call_new_api.rb
```diff
@@ -0,0 +1,23 @@
+require "test_helper"
+
+class CallTest < Minitest::Spec
+  Song = Struct.new(:title)
+
+  class SongForm < TestForm
+    property :title
+
+    validation do
+      params { required(:title).filled }
```
rforte: @emaglio I apologize in advance if this is not the right place for a discussion or if I'm missing something in the implementation of this PR. A month ago I patched reform to support dry-v 1.x and ended up with the same api and similar implementation. If this PR takes the values AFTER validation and syncs them back to the form then you can stop reading. Otherwise, it might be worthwhile calling out the behavior of using `params` vs. `schema` (and possibly surprising behavior).

With dry-v 1.x a `params` schema will perform coercion if used in conjunction with a type spec but the coerced value is not written back to the form or model. For example:
```
Foo = Struct.new(:age)

class SomeForm < TestForm
  property :age

  validation do
    # XXX: age will be coerced to an integer if possible
    params { required(:age).value(:integer) }

    rule(:age) { key.failure('value exceeded') if value > 999 } 
  end
end

model = Foo.new
form = SomeForm.new(model)
# age is probably expected to be an integer but passing a string
form.validate(age: '99') # true
form.sync
# probably not what the user expected in their model
model # <struct Foo age="99">
```
This might be surprising to users since the behavior of dry-v 1.x's Dry::Validation::Contract `params` schema writes the coerced value to the result hash. Switching from `params` to `schema` catches this issue during the type spec as would using `params {  required(:age).filled(:int?) }` but it's all rather moot. 

IMHO, while it's awesome to have dry-v 1.x support in reform 2.x the behavior will appear inconsistent to users who aren't familiar with how reform works. It's probably way more effort than it's worth to try and change anything other than documenting this behavior. 

It's highly likely there's a gap in my understanding of this PR but I'd be curious to know if there are reasons to use `params` over `schema` when the coercion support that `params` provides is limited to the context of validation. Maybe it's possible to use the coerced values in a rule to further munge data and call form setters to sync them back but that seems gross and hacky.

Thanks

emaglio: @rforte I haven't changed the way `reform` works and it basically calls the validation block as it was doing for dry-v < 1.x - regarding the coercion I believe that is up to the user to either use `schema` or `params` to get the correct values. I'm going to add some tests for it and see if what I'm saying makes sense :)

emaglio: @rforte added a commit here https://github.com/trailblazer/reform/pull/496/commits/7721562f010e738ece5ce31255426097fd39d4b5 where I have added some tests using either `schema` or `params` and I feel like reform does not need to add any "new" features if dry-v changed the way things are handled is up to the user to make sure that everything works correctly.
Hope this make sense.

rforte: @emaglio I'm sorry if it sounded like something should be changed. That wasn't my intention. All I was bringing up is that this is a functional change that should be documented to avoid programmers running into unexpected behavior based on the way dry-v 1.x works on its own. I totally agree that it's up to the end user to understand how things work but good documentation goes a long way towards understanding.

I <3 all the dry-* and trb gems and I want to see those projects succeed because they enable writing better and more understandable code. It's clear you folks are more than capable of making good technical/design decisions so keep doing what you do. Where I think the community could really benefit is clearer documentation. For example, small changes in the documentation for populators/pre-populators could easily reduce the amount of confusion they cause as evident of forum posts. I also can't be the only one that's constantly looking at these gem's spec tests to help me understand behavior that is not clearly documented or not documented at all. \</soapbox>

Anyways, thanks again for responding. Cheers!

emaglio: @rforte you are 100% right, adding a note in our Trello board to add these changes in our docs -a new website is in working in progress and it will include these changes.
Thanks again for your contribution to this :+1:

rforte: @emaglio Thanks for your time and patience. :)

PR #503, opened 2019-11-17
Drop dry-v compatibility < 1.x

emaglio:
First step to v 3.x, drop compatibility with dry-v < 1.x


emaglio:
We are going to release a `2.4` where we drop the dry-v compatibility for older versions

PR #505, opened 2020-05-20
test ruby 2.7

seuros:


.travis.yml
```diff
@@ -7,20 +7,10 @@ rvm:
   - 2.6
   - 2.5
   - 2.4
-env:
```
emaglio, line 8: I'm not sure about this..it's definitely way cleaner and faster in Travis but we don't run the test for each dry-v version which can be good since we can keep track of compatibility issue with a specific one.
What do you think?

seuros: Those version are patch version of each other. 
There is api change only between 0.1.3 and 1.0.0 and soon 2.0.0

Rakefile
```diff
@@ -10,8 +11,7 @@ TEST_WITH_OLD_AND_NEW_API = %w[
 ].freeze
 
 def dry_v_test_files
-  dry_v_version = ENV.fetch("DRY_VALIDATION", "~> 0.13.0")
-  api = dry_v_version.gsub("~>", "").to_f >= 1.0 ? "new" : "old"
+  api = Gem::Version.new(Dry::Types::VERSION).to_s.split('.').first.to_i >= 1 ? "new" : "old"
```
yogeshjain999, line 14: Do we need to check `Dry::Types` or `Dry::Validation` version here ?

emaglio: It’s the same - this is an horrible hack that I have done to keep both old/new api test and compatibility which we will drop soon. No need to spend any more time on it - happy if @seuros wants to change it to dry_validation but yeah not a big deal 👍🏼

PR #517, opened 2020-07-08
merge 2.3 into master

seuros:


test/form_option_new_api.rb
```diff
@@ -17,7 +17,7 @@ class AlbumForm < TestForm
 
   it do
     form = AlbumForm.new(Album.new(Song.new("When It Comes To You")))
-    form.song.title.must_equal "When It Comes To You"
+    _(form.song.title).must_equal "When It Comes To You"
```
emaglio, line 20: maybe let's have `assert_equal` so all the tests have the same style?

PR #518, opened 2020-07-09
rename schema to contract as we passing a contract

seuros:


saiqulhaq:
is it compatible with `2.2.x` version?
for 2.3.0, it says 
image

seuros:
it only apply if you upgraded to dry-validation 1.0+.
2.2.0 supported only 0.13 if i recall correctly. with that api, you still using need to use `schema`

PR #525, opened 2021-02-25
[~] fix memory leak when use "dry-validation"

skcc321:
The issue was detected in ~~two~~ one places on each validation call:
1.    initialization of dry-validation contract
2.   ~~initialization new array with constants inside~~

to reproduce:
1) install gem https://github.com/SamSaffron/memory_profiler
2) go to the repository and profile memory consumption by ruby script "mem_leak.rb".
```bash
ruby-memory-profiler --pretty -m 5 mem_leak.rb
```

```ruby
# mem_leak.rb
require_relative "lib/reform"
require_relative "lib/reform/form/dry"

class TestForm < Reform::Form
  feature Reform::Form::Dry
end

Song = Struct.new(:title)
class Contr < TestForm
  validation do
    option :form
    params do
      required(:title).filled(min_size?: 2)
    end

    rule do
      if form == "hello"
        base.failure('creating events is allowed only on weekdays')
      end
    end
  end
end

form = Contr.new(Song.new(nil))

500.times do
  form.(title: "hello")
end
GC.start
```
3) download this PR and run it again.
---
here are results:

```
# before fix
{1:42}[2.7.2]~/reform:master ✗ ➭ ruby-memory-profiler --pretty -m 5 mem_leak.rb                                                                                                                       
Total allocated: 29.77 MB (261957 objects)                                                                                                                                                                                    
Total retained:  29.76 MB (261895 objects)  
...
# after fix
{1:42}[2.7.2]~/reform:master-fix-memory-leak ✗ ➭ ruby-memory-profiler --pretty -m 5 mem_leak.rb                                                                                                                 
Total allocated: 2.12 MB (20693 objects)                                                                                                                                                                                      
Total retained:  2.11 MB (20624 objects)
...
```

as a bonus, performance results:
```
# before
       user     system      total        real
   1.708815   0.072268   1.781083 (  1.782192)
# after 
       user     system      total        real
   0.181692   0.000599   0.182291 (  0.182303)
```

### DIFF
--- a/lib/reform/form/dry.rb
+++ b/lib/reform/form/dry.rb
@@ -46,8 +46,11 @@ module Reform::Form::Dry
 
         dynamic_options = { form: form }
         inject_options = schema_inject_params.merge(dynamic_options)
+        contract.new(inject_options).call(input_hash(form))
+      end
 
-        validator.build(inject_options, &block).call(input_hash(form))
+      def contract
+        @contract ||= Class.new(validator, &block)
       end
     end
   end
--- a/test/module_test.rb
+++ b/test/module_test.rb
@@ -25,7 +25,7 @@ class ModuleInclusionTest < MiniTest::Spec
       params { required(:band).filled }
     end
 
-    include Dry::Types.module # allows using Types::* in module.
+    include Dry.Types(default: :nominal) # allows using Types::* in module.
     property :cool, type: Types::Params::Bool # test coercion.
   end
 


skcc321:
@seuros @emaglio could you guys take a look at that?

apotonick:
Thanks, this is great work!

I wonder why there are object sticking around in the first place. Have you encountered this without dry-validation, too?

skcc321:
> Thanks, this is great work!
> 
> 
> 
> I wonder why there are objects sticking around in the first place. Have you encountered this without dry-validation, too?

As a workaround, we switched to ActiveModel validation on some services on production: image
But we have other services where we still have dry + reform. That is why I started debug on dry contract first, but during investigation found not garbage collected arrays with references to the same constants.

I'll try to check it without "dry" soon.

apotonick:
Well, I'm happy to merge that fix if it's urgent?

seuros:
Im testing it

ybregey (approved):

skcc321:
OK. Looks like I'm wrong about that array initialization. GC can garbage collect it. Initially, I tried to reproduce it inside tests, and I noticed it using that setup. It is not reproducible with the current setup. So looks like that is not the case. I'll remove that part from the PR.

serhii-bodnaruk (approved):

skcc321:
any updates on that? @seuros

seuros:
@skcc321  I think we are good. I found another memory leak, but it seem that it coming from dry-validation. I will write a report soon.

skcc321:
@seuros if so, how soon it can be released?

seuros:
This weekend for sure.

seuros:
@skcc321  on rubygems.

skcc321:
@seuros thank you!

PR #536, opened 2022-05-17
Loosen Representable dependency to allow versions < 4

richardboehme:
This allows us to use representable 3.2.0 and other new minor versions.


apotonick (approved):

richardboehme:
@apotonick Do you know why the CI fails with Ruby head? Did they introduce some changes that are incompatible with reform? I do not think it relates with my change.

apotonick:
@richardboehme Looks like a `pry` problem, remove it from the Gemfile if you feel like it. I never use it anyway.

